PR #6, opened 2021-01-27
Home Screen Book List

Anshul-Birla:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Added "New Books" section
- Added "All Books" section
- Added support for custom font (Gotham Rounded)
- Added circle at the top of page

### Testing
- Rendered on both Anshul and Aksharan's screens

### Confirmation of Change 
Main Screen Image


Anshul-Birla:
Did not add comments to the code. Had some questions we wanted to ask you in Wednesday's meeting, but thought you would like to see what we have done so you could give us pointers tomorrow also :)

mobile/src/components/BookCard.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { Image, StyleSheet, View } from 'react-native';
+import { Book } from '../models/Book';
+
+type BookProps = { book: Book, index: number };
+
+export const BookCard: React.FC<BookProps> = ({ book, index }) => {
+  return (
+    <View style={styles.container}>
+      <View style={styles.shadow}>
```
dacruz21, line 10: Can these not be combined into one View?

mobile/src/components/BookList.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { FlatList, View } from 'react-native';
+import { Book } from '../models/Book';
+import { BookCard } from './BookCard';
+
+type BookListProps = {books: Book[] };
+
+export const BookList: React.FC<BookListProps> = ({ books }) => {
```
dacruz21, line 8: Maybe `HorizontalBookList` for this component, if we're renaming AllBooks too?

mobile/src/models/Book.ts
```diff
@@ -0,0 +1,12 @@
+import { TabContent } from './TabContent';
+
+export type Book = {
+    'id': string
+    'title': string
+    'image'?: string
+    'read'?: TabContent
+    'explore'?: TabContent
+    'learn'?: TabContent
+    'created_at': string
```
dacruz21, line 10: These don't need to be quoted

mobile/src/models/TabContent.ts
```diff
@@ -0,0 +1,4 @@
+export type TabContent = {
+    video?: string;
+    body: string;
+};
```
dacruz21, line 4: Can we move this to Book.ts? I think it makes sense to keep them together - we're never going to reuse this type for anything other than books.

mobile/src/screens/HomeScreen.tsx
```diff
@@ -0,0 +1,50 @@
...
+  text: {
+    ...TextStyles.h3,
+    paddingTop: 33,
+    paddingBottom: 19,
+    paddingLeft: 17,
+  },
+  allBooks: {
+    paddingLeft: 17,
+    paddingRight: 17,
+  },
```
dacruz21, line 49: Shouldn't these properties be specified on the `<View>`s containing the `<Text>`s, not the `<Text>`s themselves? That would also give you the advantage of directly assigning `style` to `TextStyles.h3`, which makes it a little more readable IMO

mobile/src/styles/TextStyles.ts
```diff
@@ -0,0 +1,21 @@
+import { StyleSheet } from 'react-native';
+
+export const TextStyles = StyleSheet.create({
```
dacruz21, line 3: Nice!

mobile/App.tsx
```diff
@@ -1,21 +1,23 @@
...
-  container: {
-    flex: 1,
-    alignItems: 'center',
-    justifyContent: 'center',
-  },
-});
-
-// eslint-disable-next-line import/no-default-export
 export default App;
+// eslint-disable-next-line import/no-default-export
```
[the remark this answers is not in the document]
dacruz21: oh it did lol

typokign (changes requested):
Great job guys! Here are some of the easily-fixable things I noticed. Also want to go over some of the styling tomorrow - there's a lot of math in your stylesheets which is actually pretty impressive but I'm not sure is actually necessary.

dacruz21:
I'm also getting an error: "VirtualizedLists should never be nested inside plain ScrollViews with the same orientation - use another VirtualizedList-backed container instead."

Sounds like it doesn't like that the entire home screen is a ScrollView, with a FlatList inside of it. That sucks, maybe the root of the home screen can be a <FlatList> with [ListHeaderComponent](https://docs.expo.io/versions/latest/react-native/flatlist/#listheadercomponent) containing the header, horizontal book scrollers, and text headers?

typokign (dismissed):
Great work guys

approved

rohanputhukudy:
I think there's a problem with the way we're displaying all the books using ColumnBookList. I ran the app on a few different devices and found that there wasn't much space between books on the same row for smaller devices (ex. iPhone 12 mini)

Simulator Screen Shot - iPhone 12 mini - 2021-02-03 at 22 31 13

mobile/src/components/ColumnBookList.tsx
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Renders a vertical list of books with 3 books per row.
+ */
+export const ColumnBookList: React.FC<ColumnBookListProps> = ({ books }) => {
+  return (
+    <FlatList
+      data={books}
+      renderItem={({ item }) => (
+        <View style={styles.bookCard}>
+          <BookCard book={item} size={120} />
```
rohanputhukudy, line 17: I think having the size always be 120 is what's causing the issue with spacing on devices with smaller screen sizes. Changing this to depend on the screen resolution may be better. However, this would require using the dimensions of the device to calculate image size since the book image comes from a url and so we have to specify the exact dimensions of the image.

rohanputhukudy:
Currently, if the total number of books is congruent to 2 modulo 3, the last row of the all books grid ends up formatted a bit weirdly.

Simulator Screen Shot - iPhone 12 Pro Max - 2021-02-04 at 00 04 01

Anshul-Birla:
> Currently, if the last row of the all books grid has 2 books then it ends up formatted a bit weirdly.
> 
> 
> 
> Simulator Screen Shot - iPhone 12 Pro Max - 2021-02-04 at 00 04 01

This bug is known and we are going to fix it in a later version

dacruz21:
Yep, I want to check with design to see if those last 2 should be left-aligned, or if they should be spaced evenly, or something else. So I'm fine with just living with it for now.

mobile/src/components/ColumnBookList.tsx
dacruz21: Damn. @Anshul-Birla @aksharans maybe we should use `Dimensions` in this case since we're always showing 3 books in a row. Maybe something like `width * 0.25` so it's a little less than 1/3 to give room for spacing? Hopefully that looks good on most screen sizes.

aksharans: I think that should be fine, but if we are doing size based on the dimensions of the device, there could be a chance that those images look smaller compared to the New Books for You images, so I think it's just a matter if the design is okay with not having the All Books images being larger than the other booklist images and being a set size.

dacruz21: I think that's fine. Main idea here is that density independent pixels probably aren't going to work when we're not in a scrolling container.

rohanputhukudy (dismissed):
This looks good to me now

typokign (approved):

rohanputhukudy (approved):

PR #7, opened 2021-02-02
Backend Books API

k6han:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Implemented backend api for getting books
- Added route /books to pull list of (id, title, author, image, created_at)
- Added route /books/{id} to pull full contents of a book

### Testing
- Made a testing file, ran basic tests using a golang httptest server

### Confirmation of Change 

Run make test in words-alive/backend


backend/internal/controllers/book.go
```diff
@@ -0,0 +1,41 @@
+package controllers
+
+import (
+	"encoding/json"
+	"github.com/go-chi/chi"
+	"net/http"
+
+	"github.com/TritonSE/words-alive/internal/database"
+	//"github.com/TritonSE/words-alive/internal/models"
```
dacruz21, line 9: Cruft, can you delete this?

backend/internal/controllers/book_test.go
```diff
@@ -0,0 +1,86 @@
...
+// Test for the book list function
+func TestGetBooks(t *testing.T) {
+	// Populate database
+	conn.Exec("INSERT INTO books (title, author) values ('c','c1');")
+	conn.Exec("INSERT INTO books (title, author) values ('a','a1');")
+	conn.Exec("INSERT INTO books (title, author) values ('b','b1');")
+
+	// Open server
+	ts := httptest.NewServer(r)
+	defer ts.Close()
```
dacruz21, line 37: Can you move these to TestMain? That way we can recycle the same test server for each module.

backend/internal/controllers/book_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	var response []models.Book
+	defer res.Body.Close()
+
+	// Parse response
+	body, err := ioutil.ReadAll(res.Body)
+	require.NoError(t, err)
+
+	err = json.Unmarshal(body, &response)
+	require.NoError(t, err)
```
dacruz21, line 52: This is a lot of boilerplate for something we'll be doing in almost every test. Can you through together a quick testutils package that does all of this? The return type can be `interface{}` and then you can just typecast to the expected type in each test.

backend/internal/controllers/book_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Check for correct number of elements, and sort alphabetically
+	require.Len(t, response, 3)
+	require.Equal(t, response[0].Title, "a")
+	require.Equal(t, response[1].Title, "b")
+	require.Equal(t, response[2].Title, "c")
```
dacruz21, line 58: Awesome, glad we're testing for sort order :)

backend/internal/models/book.go
```diff
@@ -0,0 +1,29 @@
...
+	ID            string             `json:"id"`
+	Title         string             `json:"title"`
+	Author        string             `json:"author"`
+	Image         *string            `json:"image"`         // image link
+	Read_Video    *string            `json:"read_video"`    // youtube link
+	Read_Body     *string            `json:"read_body"`     // markdown
+	Explore_Video *string            `json:"explore_video"` // youtube link
+	Explore_Body  *string            `json:"explore_body"`  // markdown
+	Learn_Video   *string            `json:"learn_video"`   // youtube link
+	Learn_Body    *string            `json:"learn_body"`    // markdown
```
dacruz21, line 18: This doesn't match the JSON schema the frontend is already expecting. Can we add:

```go
type TabContent struct {
    Video string `json:"video"`
    Body string `json:"body"`
}
```

and include them here like so:

```go
    Read TabContent `json:"read"`
    ...
```

backend/internal/models/book.go
```diff
@@ -0,0 +1,29 @@
...
+	Created_At    pgtype.Timestamptz `json:"created_at"`    // Following ISO 8601
+}
+
+// For main page listings
+type Book struct {
+	ID         string             `json:"id"`
+	Title      string             `json:"title"`
+	Author     string             `json:"author"`
+	Image      *string            `json:"image"`      // image link
+	Created_At pgtype.Timestamptz `json:"created_at"` // Following ISO 8601
```
dacruz21, line 28: s/Created_At/CreatedAt please

Also, pgx should be able to scan into a `time.Time`, which is more idiomatic than having Postgres-specific types in our models.

backend/internal/controllers/book.go
```diff
@@ -0,0 +1,41 @@
...
+
+type BookController struct {
+	Books database.BookDatabase
+}
+
+// Fetches a list of all books sorted by title for the main page (/books)
+func (c *BookController) GetBookList(rw http.ResponseWriter, req *http.Request) {
+	books, err := c.Books.FetchBookList(req.Context())
+	if err != nil {
+		rw.WriteHeader(500)
```
dacruz21, line 20: I think it might be a good idea to use the `http` package status code constants, just since not all of us have them memorized yet :) https://golang.org/pkg/net/http/#pkg-constants

Can you replace 500 with http.StatusInternalServerError?

typokign (changes requested):
Looking good guys, thanks for documenting things and writing tests :D 

Let's try to cut some of the boilerplate with helper functions though. I left a comment in the tests, but I also had this idea:

Let's write a `func writeResponse(rw http.ResponseWriter, statusCode int, body interface{})` which does a `rw.WriteHeader` followed by a `json.NewEncoder(rw).Encode(body)`. You can put this in a `utils.go` or something in the controllers package, so it's available to all of the controllers.

Thanks!

backend/internal/models/book.go
dacruz21: Also run `go mod tidy` and `go mod vendor` after doing this, since you won't need any of the vendored pgtype code

backend/internal/database/book.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// Define our list of books to return
+	books := make([]models.Book, 0)
+
+	// Loop over all of the rows that were returned (one for each book in the database)
+	for rows.Next() {
+		// Create a new book object, then "scan" the column values from the row into the struct.
+		var book models.Book
+		// Pay attention to the order, since we did SELECT id, title, we need to scan the ID before the Title
+		if err := rows.Scan(&book.ID, &book.Title, &book.Author, &book.Image, &book.Created_At); err != nil {
```
Anshul-Birla, line 36: maybe split this up into 2 lines?

backend/internal/models/book.go
Anshul-Birla, line 28: The frontend is expecting "CreatedAt" for the structure.

backend/internal/database/book.go
```diff
@@ -0,0 +1,64 @@
...
+
+	defer rows.Close()
+
+	// Define our list of books to return
+	books := make([]models.Book, 0)
+
+	// Loop over all of the rows that were returned (one for each book in the database)
+	for rows.Next() {
+		// Create a new book object, then "scan" the column values from the row into the struct.
+		var book models.Book
```
Anshul-Birla, line 34: Don't really think we need all these comments

backend/internal/models/book.go
dacruz21: I think we want snake_case for JSON fields - we can resolve the discrepancy when we connect the app to the API.

k6han: Should we keep Created_At for now then?

dacruz21: The field should be named `CreatedAt` but the JSON tag should be `json:"created_at"`

backend/internal/controllers/book.go
```diff
@@ -0,0 +1,37 @@
...
+
+type BookController struct {
+	Books database.BookDatabase
+}
+
+// Fetches a list of all books sorted by title for the main page (/books)
+func (c *BookController) GetBookList(rw http.ResponseWriter, req *http.Request) {
+	books, err := c.Books.FetchBookList(req.Context())
+	if err != nil {
+		WriteResponse(rw, http.StatusInternalServerError, books)
```
dacruz21, line 18: `books` here is probably going to be empty string, or something invalid. Just replace `books` with `"error"` for now.

backend/internal/controllers/book.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Fetches all contents of a book for reading (/book/{id})
+func (c *BookController) GetBookDetailsByID(rw http.ResponseWriter, req *http.Request) {
+
+	var bookID string = chi.URLParam(req, "id")
+
+	bookDetails, err := c.Books.FetchBookDetailsByID(req.Context(), bookID)
+	if err != nil {
+		WriteResponse(rw, http.StatusInternalServerError, bookDetails)
```
dacruz21, line 32: ^ same

backend/internal/controllers/utils.go
```diff
@@ -0,0 +1,12 @@
+package controllers
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// Sends data and status code for http
+func WriteResponse(rw http.ResponseWriter, statusCode int, body interface{}) {
```
dacruz21, line 9: You can lowercase this to writeResponse, all the controllers are in the same package.

backend/internal/models/book.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Contains the contents of a book (all book information)
+type BookDetails struct {
+	ID        string      `json:"id"`
+	Title     string      `json:"title"`
+	Author    string      `json:"author"`
+	Image     *string     `json:"image"` // image link
+	Read      *TabContent `json:"read"`
+	Explore   *TabContent `json:"explore"`
+	Learn     *TabContent `json:"learn"`
```
dacruz21, line 22: These should always exist, so we don't need to make them pointers.

backend/internal/models/book.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"time"
+)
+
+type TabContent struct {
+	Video *string `json:"video"`
+	// TODO once we have real testing data make Body a string.
+	// Leave as *string for now to pass unit tests
+	Body *string `json:"body"`
```
dacruz21, line 11: We also want this to always exist, so make this a string please.

backend/internal/testutils/testutils.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"encoding/json"
+	"github.com/stretchr/testify/require"
+	"io/ioutil"
+	"net/http"
+	"testing"
+)
+
+// Sends http request, converts data from json, and stores to response
+func SendHttpRequest(req *http.Request, response interface{}, t *testing.T) {
```
dacruz21, line 12: IMO crafting request objects is a pain in the ass. It would be awesome if we could write most tests in like 2-3 lines.

maybe `func MakeRequest(t *testing.T, ts httptest.Server, method string, url string, response interface{})`?

dacruz21: Also, then you can concatenate `ts.URL` with `url` within the function.

backend/internal/database/book.go
```diff
@@ -0,0 +1,74 @@
...
+		&read.Body, &explore.Video, &explore.Body,
+		&learn.Video, &learn.Body, &book.CreatedAt)
+
+	if err != nil {
+		return book, errors.Wrap(err, "error on SELECT FROM books in FetchBookByID")
+	}
+
+	book.Read = &read
+	book.Explore = &explore
+	book.Learn = &learn
```
dacruz21, line 71: Once you get rid of the pointers in the models, you can get rid of the `read, explore, learn` vars, and scan into `&book.Read.Video`, `&book.Read.Body` etc.

typokign (changes requested):

typokign (approved):
Great work!

lgtm

Anshul-Birla (approved):
Sorry for the delay, but LGTM :)

PR #13, opened 2021-02-17
Mobile API client and backend integration

Anshul-Birla:
### Administrative Info
Monday Board ID: 1021230873

### Changes
What changes did you make?
- Added dynamic url capability 
- Added activity indicator for when books do not load 

### Testing
- Ran on both Aksharan and Anshul's phone

### Confirmation of Change 
- Use export BASE_URl = <domain_name> and then run npm start to see the domain specified is the domain axios connects to 
- Run npm run dev to see that localhost:8080 is the default url for axios connection


mobile/src/screens/HomeScreen.tsx
```diff
@@ -19,6 +19,19 @@ export const HomeScreen: React.FC = () => {
     .slice(0, 5);
   const allBooks = booksCtx.books;
 
+  let newBooksRender;
+  let colBooksRender;
+  const booksLoaded: boolean = newBooks.length > 0;
```
dacruz21, line 24: This feels kinda hacky. Can we add a `loading` bool to the `BookContext`/`BookState` type? That's a pretty common pattern in React data fetching. The idea being that `fetchBooks()` would set `loading` to true, fetch the books, then set `loading` to false. Then you can just do `booksCtx.loading`.

mobile/src/screens/HomeScreen.tsx
```diff
@@ -19,6 +19,19 @@ export const HomeScreen: React.FC = () => {
...
+
+  // only load books if array is non-empty. Else load acitivty indicator
+  if (booksLoaded) {
+    newBooksRender = <HorizontalBookList books={newBooks} />;
+    colBooksRender = <ColumnBookList books={allBooks} />;
+  } else {
+    newBooksRender = <ActivityIndicator/>;
+    colBooksRender = <ActivityIndicator/>;
+  }
```
dacruz21, line 33: This isn't bad but if you want to make it a little less verbose you can just do:

```
{ booksCtx.loading && <ActivityIndicator/> }
{ !booksCtx.loading && <HorizontalBookList books={newBooks} }
```

within your JSX.

mobile/src/context/BookContext.tsx
```diff
@@ -15,157 +15,18 @@ const initialState: BookState = {
...
-      },
-
-    ];
-    dispatch({ type: 'BOOKS_LOADED', payload: books });
+    client.getBooks().then((res) => {
+      dispatch({ type: 'BOOKS_LOADED', payload: res });
+    }).catch(() => {
+      const empty: Book[] = [];
+      dispatch({ type: 'BOOKS_LOADED', payload: empty });
+    });
```
dacruz21, line 29: I think in the error case we should just log a message to the console and do nothing (leave loading as `true` and don't set books to an empty array). When we set up offline mode we can have the catch statement trigger the "You are offline, functionality limited" view. No need to do that now though.

Also can you convert this to async/await?

mobile/src/context/APIContext.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { createContext } from 'react';
+import { WordsAliveAPI } from '../classes/WordsAliveAPI';
+
+//set baseURL using environment varibales
+const baseURL = 'http://' + (process.env.BASE_URL || 'localhost:8080');
```
dacruz21, line 5: The protocol should be part of the BASE_URL, since staging and prod are going to be over https. You can just do `const baseURL = process.env.BASE_URL || 'http://localhost:8080';`

dacruz21: Also I thought we figured out we need to do this? https://docs.expo.io/guides/environment-variables/#using-app-manifest--extra What changed?

mobile/.env
```diff
@@ -0,0 +1 @@
+BASE_URL=https://localhost:8080
```
dacruz21, line 1: I don't think we need a .env file for this, it's only one variable. I'd rather just specify the variable in the npm commands.

mobile/babel.config.js
```diff
@@ -2,5 +2,7 @@ module.exports = function(api) {
   api.cache(true);
   return {
     presets: ['babel-preset-expo'],
+    plugins: ['transform-inline-environment-variables'],
```
dacruz21, line 5: I think this is only necessary for .env file support, so we can remove it with the .env file.

mobile/package.json
```diff
@@ -4,18 +4,22 @@
...
     "@react-navigation/bottom-tabs": "^5.11.7",
     "@react-navigation/native": "^5.9.2",
+    "babel-plugin-inline-dotenv": "^1.6.0",
     "expo": "^40.0.0",
     "expo-app-loading": "^1.0.1",
+    "expo-font": "~8.4.0",
     "react": "16.13.1",
     "react-dom": "16.13.1",
     "react-native": "^0.63.4",
+    "react-native-dotenv": "^2.5.3",
```
dacruz21, line 14: same

mobile/package.json
```diff
@@ -24,6 +28,7 @@
     "@types/react-native": "~0.63.2",
     "@typescript-eslint/eslint-plugin": "^4.10.0",
     "@typescript-eslint/parser": "^4.10.0",
+    "babel-plugin-transform-inline-environment-variables": "^0.4.3",
```
dacruz21, line 31: and same

mobile/package.json
```diff
@@ -40,7 +45,8 @@
     "android": "expo start --android",
     "ios": "expo start --ios",
     "web": "expo web",
-    "eject": "expo eject"
+    "eject": "expo eject",
+    "dev": "unset BASE_URL && npm start"
```
dacruz21, line 49: I'd rather this be more explicit, as `"dev": "BASE_URL=http://localhost:8080 npm start"`

typokign (changes requested):
Good start, see below

mobile/src/context/APIContext.tsx
Anshul-Birla: Turns out for programs to have access to env variables, you need to export them which neither me or Aksharan were doing. If we use the babel config, we can just use process.env. No app.config needed

mobile/src/context/BookContext.tsx
Anshul-Birla: ah should we disable the linter for that line then? We had a console.log there but the linter complains

mobile/.env
Anshul-Birla: yeah this should not have made it in the commit. I think we just forgot to delete it

mobile/src/context/BookContext.tsx
dacruz21: ugh let's just turn that rule off for now, we can clean up the console.logs later if we come up with a better logging pipeline.

"no-console": "off" in eslintrc

mobile/src/context/APIContext.tsx
dacruz21: The problem is:

> Unlike .env, this property is included when you publish your project with expo publish or expo build

When we build the app as a standalone APK/ISA this will all fall apart. If we reference environment variables in app.config.js, then:

> The manifest is loaded when starting or publishing your app and has access to the environment variables using process.env

mobile/babel.config.js
Anshul-Birla: We need it if we do not have the app.config file. The reason Aksharan and I did not want to use it is because in the EXPO documentation, it says that app.config.ts is experimental and can change at anytime, so we thought it was best to avoid it

dacruz21: Referring to this? https://docs.expo.io/workflow/configuration/#using-typescript-for-configuration-appconfigts-instead-of

Let's just use app.config.js then, it's just build configuration, type safety isn't as big a deal

Anshul-Birla: Yup. No worries then, we can get that change done quickly. Will resolve all of the babel stuff too.

mobile/package.json
```diff
@@ -42,7 +44,8 @@
     "android": "expo start --android",
     "ios": "expo start --ios",
     "web": "expo web",
-    "eject": "expo eject"
+    "eject": "expo eject",
+    "dev": "export BASE_URL=http://localhost:8080 && npm start"
```
dacruz21, line 48: This should just be `BASE_URL=http://localhost:8080 npm start`, no export or &&.

mobile/src/classes/WordsAliveAPI.tsx
```diff
@@ -0,0 +1,18 @@
...
+      this.client = axios.create({ baseURL: baseURL });
+    }
+
+    async getBooks(): Promise<Book[]> {
+      const res = await this.client.get('/books');
+      return res.data;
+    }
+}
+
+export { WordsAliveAPI };
```
dacruz21, line 18: Can we move this to a `src/api/` dir instead of classes? Should make it easier to find and extend if we want to. Also can you document getBooks()? Lastly can you indent this to 2 spaces? No idea why the lint didn't pick that up.

mobile/src/context/BookContext.tsx
```diff
@@ -1,191 +1,57 @@
...
 type BookState = {
     books: Book[], // holds all the books we have stored
     fetchBooks: () => void, // api call to get books
+    loading: boolean, // lets us know if the books are being loaded or not
 };
 
 const initialState: BookState = {
   books: [],
   fetchBooks: () => {},
+  loading: true,
```
dacruz21, line 14: Technically the context is _not_ loading anything until the first call to fetchBooks(), so this should be false.

mobile/src/context/BookContext.tsx
```diff
@@ -1,191 +1,57 @@
...
-      },
-
-    ];
-    dispatch({ type: 'BOOKS_LOADED', payload: books });
+    client.getBooks().then((res) => {
+      dispatch({ type: 'BOOKS_LOADED', payload: res, loading: false });
+    }).catch((err) => {
+      console.log(err);
+      // if books not found, continue indicating loading
+      dispatch({ type: 'ERROR', payload: null, loading: true });
```
dacruz21, line 30: I don't think we need to dispatch at all here, we can just keep the current state of `loading: true`. This just adds another state to the books state machine (and believe me, it's going to get messy once we add offline mode).

mobile/src/context/BookContext.tsx
```diff
@@ -1,191 +1,57 @@
...
-        id: 'x',
-        title: 'bar',
-        image: 'https://placekitten.com/200/305',
-        createdAt: '1000',
-      },
-
-    ];
-    dispatch({ type: 'BOOKS_LOADED', payload: books });
+    client.getBooks().then((res) => {
+      dispatch({ type: 'BOOKS_LOADED', payload: res, loading: false });
```
dacruz21, line 26: Typically dispatch only takes a `type` and a `payload`. `payload` should be reserved for extra data, like API responses. Instead, you should set the `loading` bool within the reducer.

mobile/src/context/BookContext.tsx
```diff
@@ -1,191 +1,57 @@
...
+        loading: state.loading,
       }}
     >
       {children}
     </BookContext.Provider>
   );
 };
 
-type BookAction = {type: 'BOOKS_LOADED', payload: Book[]};
+type BookAction = {type: string, payload: Book[], loading: boolean };
```
dacruz21, line 47: You'll want to make the `type` a union so we can keep type safety, and drop the loading boolean.

typokign (changes requested):
Thanks for fixing merge conflicts. I'll ping you guys on Slack so we can talk a little more about how reducers are supposed to work. Also can you regen package-lock.json? It got deleted somehow.

Anshul-Birla:
> Thanks for fixing merge conflicts. I'll ping you guys on Slack so we can talk a little more about how reducers are supposed to work. Also can you regen package-lock.json? It got deleted somehow.

Uh for some reason, github keeps on saying the file is deleted on the PR, but if you look at the actual file, its there

mobile/src/screens/HomeScreen.tsx
```diff
@@ -52,15 +52,15 @@ export const HomeScreen: React.FC = () => {
       </View>
 
       <View>
-        <HorizontalBookList books={newBooks} />
+        { booksCtx.loading ? <ActivityIndicator/> : <HorizontalBookList books={newBooks}/> }
```
dacruz21, line 55: The default props for `<ActivityIndicator>` aren't visible on Android :(

I played with it a little on my phone and found that `<ActivityIndicator size="large" color={Colors.orange}/>` looked best on Android, but I don't know if that would look good on iOS too.

Can you guys add a `<LoadingCircle>` component to `src/components`, wrapping an ActivityIndicator for consistency across the app? If size="large" and color=Colors.orange looks good on iOS you can just drop that right into the return statement, but if some other color (or the default color) looks better you can use the [Platform module](https://reactnative.dev/docs/platform-specific-code) to customize it for each OS. Definitely make it `size="large"` on iOS though for consistency.

typokign (changes requested):
Sorry, I know you're super busy this week Anshul, but this should be a real quick fix.

typokign (approved):
approved

allisonbhavsar (approved):
Axios connection working on my device and books are rendering from the db, lgtm!

PR #14, opened 2021-02-17
Support multiple languages in books API

k6han:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
Refactored the database to include support for multiple languages, and refactored the routing to work with the new database.

### Testing
I wrote tests in book_test.go in controllers, and verified that the new languages were being read.

### Confirmation of Change 
Run make test in the backend directory. Open internal/controllers/book_test.go to see the tests.


backend/internal/database/book.go
```diff
@@ -49,23 +51,51 @@ func (db *BookDatabase) FetchBookList(ctx context.Context) ([]models.Book, error
...
 		return book, errors.Wrap(err, "error on SELECT FROM books in FetchBookByID")
 	}
 
 	return book, nil
 }
+
+/*
+ * Check for existence of a book with the specified id
+ */
+func (db *BookDatabase) CheckBookID(ctx context.Context, id string,
```
dacruz21, line 87: Is there any reason you added this func instead of just making FetchBookDetailsById return a pointer?

I think this would be useful if there was any situation where we wanted to check for the existence of a book *without* fetching the book details right after, but I don't think we'll run into that situation. As of now the request needs to make 2 network calls which is just a little inefficient.

backend/migrations/20210210112014_refactor_for_lang.sql
```diff
@@ -0,0 +1,34 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE books DROP COLUMN read_video;
+ALTER TABLE books DROP COLUMN read_body;
+ALTER TABLE books DROP COLUMN explore_video;
+ALTER TABLE books DROP COLUMN explore_body;
+ALTER TABLE books DROP COLUMN learn_video;
+ALTER TABLE books DROP COLUMN learn_body;
```
dacruz21, line 8: If we had staging up already I would ask you to migrate the data from books to book_contents (with language en). But you lucked out on timing, so you don't have to do that :))))

backend/migrations/20210210112014_refactor_for_lang.sql
```diff
@@ -0,0 +1,34 @@
...
+
+CREATE TABLE book_contents(
+    id text NOT NULL,
+    lang text NOT NULL,
+	read_video text,        -- video link
+	read_body text,
+	explore_video text,     -- video link
+	explore_body text,
+	learn_video text,       -- video link
+	learn_body text,
```
dacruz21, line 18: I WILL make you reindent these though plz it hurts my eyes

backend/migrations/20210210112014_refactor_for_lang.sql
```diff
@@ -0,0 +1,34 @@
...
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE books ADD read_video text;
+ALTER TABLE books ADD read_body text;
+ALTER TABLE books ADD explore_video text;
+ALTER TABLE books ADD explore_body text;
+ALTER TABLE books ADD learn_video text;
+ALTER TABLE books ADD learn_body text;
```
dacruz21, line 31: I think this should be `ADD COLUMN`? You can test the down migration by migrating the database (by running the backend), then running `goose postgres <conn string> down`.

backend/internal/database/book.go
```diff
@@ -49,23 +51,51 @@ func (db *BookDatabase) FetchBookList(ctx context.Context) ([]models.Book, error
...
+		&book.ID, &book.Title, &book.Author, &book.Image,
+		&book.CreatedAt)
+
+	if err != nil {
+		return book, errors.Wrap(err, "error on SELECT FROM books in FetchBookByID")
+	}
+
+	err = db.Conn.QueryRowEx(ctx, queryContent, nil, id, lang).Scan(
+		&book.Read.Video, &book.Read.Body, &book.Explore.Video, &book.Explore.Body,
+		&book.Learn.Video, &book.Learn.Body)
```
dacruz21, line 75: I think you can accomplish this in 1 query with a join. Atm it adds a 3rd network call to each request, so it would be nice to consolidate them.

backend/internal/database/book.go
```diff
@@ -49,23 +51,51 @@ func (db *BookDatabase) FetchBookList(ctx context.Context) ([]models.Book, error
  * For use after a user has selected a book to read
  */
 func (db *BookDatabase) FetchBookDetailsByID(ctx context.Context,
```
dacruz21, line 53: Minor thing but `ByID` isn't quite accurate anymore, we could do `ByIDAndLang` or we could just do `FetchBookDetails` - I think the latter is concise without affecting clarity, since you kinda always need the ID and lang to get book details.

backend/internal/controllers/router.go
```diff
@@ -29,7 +30,10 @@ func GetRouter() chi.Router {
 
 	r.Route("/books", func(r chi.Router) {
 		// "localhost:8080/books/{id}
-		r.Get("/{id}", bookController.GetBookDetailsByID)
+		r.Route("/{id}", func(r chi.Router) {
+			//bookController.GetBookDetailsByID)
```
dacruz21, line 34: Dead code right here.

Also you don't need a `.Route()` call for every slash in the request, this could just be `r.Get("/{id}/{lang}", bookController.GetBookDetailsByID)`. 

You could even flatten this entire section to:

```
r.Get("/books", ...)
r.Get("/books/{id}/{lang}", ...)
```

if you wanted to.

backend/internal/controllers/book_test.go
```diff
@@ -51,8 +77,8 @@ func TestGetBooks(t *testing.T) {
 func TestGetBookDetailsByID(t *testing.T) {
 
 	var response models.BookDetails
-	testutils.MakeHttpRequest("GET", ts.URL+"/books/catcher", &response, t)
+	testutils.MakeHttpRequest("GET", ts.URL+"/books/catcher/en", &response, t)
 
 	// Check title
-	require.Equal(t, "catcher in the rye", response.Title)
+	require.Equal(t, "catcher_rb", response.Read.Body)
```
dacruz21, line 83: Can we add the title check back in too please? And maybe some more tests for:

* Book not found
* Book found, but no contents for that book in the specified lang
* Test with a book that has contents in multiple languages, just to double-check we're selecting the correct language

backend/internal/controllers/book.go
```diff
@@ -26,8 +27,21 @@ func (c *BookController) GetBookList(rw http.ResponseWriter, req *http.Request)
...
+
+	valid, err := c.Books.CheckBookID(req.Context(), bookID, lang)
+	if err != nil {
+		writeResponse(rw, http.StatusInternalServerError, "error")
+		return
+	}
+
+	// Send 404 error if book does not exist
+	if !valid {
+		writeResponse(rw, http.StatusNotFound, "error")
```
dacruz21, line 40: We can put descriptive error messages on 4xx requests, just not 500. Can this just say "book not found"?

typokign (changes requested):
Good work, thanks for taking care of the Content-Type header! You have a .DS_Store in backend, can you add that to the top-level .gitignore please?

backend/internal/controllers/utils.go
```diff
@@ -8,5 +8,6 @@ import (
 // Sends data and status code for http
 func writeResponse(rw http.ResponseWriter, statusCode int, body interface{}) {
 	rw.WriteHeader(statusCode)
+	rw.Header().Set("Content-Type", "application/json")
```
dacruz21, line 11: Actually you need to swap these two lines. `WriteHeader` *literally* writes the entire HTTP header (from `HTTP/<version> <code>` to the end of the headers) to the output stream, so you need to be done with setting headers by then.

backend/internal/models/book.go
```diff
@@ -28,4 +28,5 @@ type Book struct {
 	Author    string    `json:"author"`
 	Image     *string   `json:"image"`      // image link
 	CreatedAt time.Time `json:"created_at"` // Following ISO 8601
+	Languages []string  `json:"languages"`  // Array of languages (ISO 639-1)
```
dacruz21, line 31: really minor but can you shift this up one line? Usually created_at fields appear at the bottom of JSON objects

backend/migrations/20210210112014_refactor_for_lang.sql
dacruz21, line 8: read_body, explore_body, and learn_body need a `NOT NULL` constraint or you will get an error trying to scan a null into a `TabContent.Body` which is a string

typokign (changes requested):
Tested locally and found a few more things. Thanks!

typokign (dismissed):
Great work, thanks! Just fix the merge conflicts and this should be g2g

approved

backend/internal/controllers/book.go
```diff
@@ -23,15 +24,25 @@ func (c *BookController) GetBookList(rw http.ResponseWriter, req *http.Request)
 }
 
 // Fetches all contents of a book for reading (/book/{id})
```
Anshul-Birla, line 26: i think this comment needs updating

Anshul-Birla (changes requested):
The code itself looks good, but one small problem. Running it on my laptop, i get the error: **failed to migrate: ../../migrations directory does not exist**. I think fixing the merge conflict should fix it, because it should have the patch that David added. Once that is fixed, the branch should be g2g

dacruz21:
Just noticed the .DS_Store is still there, can you make sure that gets deleted when you fix conflicts? Thanks!

k6han:
> The code itself looks good, but one small problem. Running it on my laptop, i get the error: **failed to migrate: ../../migrations directory does not exist**. I think fixing the merge conflict should fix it, because it should have the patch that David added. Once that is fixed, the branch should be g2g

@Anshul-Birla could you test again and let me know if it works? I thought I had fixed the merge conflicts already, my bad on taking so long I had a busy week.

Anshul-Birla:
> > The code itself looks good, but one small problem. Running it on my laptop, i get the error: **failed to migrate: ../../migrations directory does not exist**. I think fixing the merge conflict should fix it, because it should have the patch that David added. Once that is fixed, the branch should be g2g
> 
> @Anshul-Birla could you test again and let me know if it works? I thought I had fixed the merge conflicts already, my bad on taking so long I had a busy week.

It still complains with the same error. Its odd because if I run the backend on main, it works perfectly fine.

dacruz21:
@k6han Merge conflicts aren't fixed yet, did you make sure to pull main before merging/rebasing?

typokign (approved):

Anshul-Birla (approved):

PR #15, opened 2021-02-24
User (family) authentication middleware and create user profile API

RohithKasar:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
Created user model, can now create users given a firebase uid
Wrote authentication middleware that ensures user has valid token before proceeding

### Testing
Attempted to create users with no token, invalid token, valid token, and duplicate users and asserted the appropriate status codes were returning
### Confirmation of Change 

run make test in backend


dhrvsood (dismissed):

dhrvsood (dismissed):

dacruz21:
Neither of you have your .gitconfig set up correctly, can you double-check the name and email address and make sure the email matches the one you use on GitHub?

image

backend/internal/auth/auth.go
```diff
@@ -0,0 +1,9 @@
+package auth
+
+import (
+	"context"
+)
+
+type Authenticator interface {
+	VerifyToken(ctx context.Context, token string) (string, bool)
```
dacruz21, line 8: Documentation on this interface and function would be useful.

backend/internal/auth/firebase.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"context"
+	"firebase.google.com/go/v4/auth"
+)
+
+type FirebaseAuthenticator struct {
+	client auth.Client
+}
+
+func (a *FirebaseAuthenticator) VerifyToken(ctx context.Context, authHeader string) (string, bool) {
```
dacruz21, line 12: Documentation on this implementation would be helpful (mention that it verifies the JWT is signed and is valid for the Firebase project)

backend/internal/auth/mock.go
```diff
@@ -0,0 +1,16 @@
+package auth
+
+import (
+	"context"
+	"strings"
+)
+
+type MockAuthenticator struct{}
+
+func (a MockAuthenticator) VerifyToken(ctx context.Context, token string) (string, bool) {
```
dacruz21, line 10: Document this with the structure of test tokens (that they are of the form `test-token<user ID>`)

backend/internal/controllers/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	conn.Exec("INSERT INTO book_contents (id, lang, read_video, read_body, " +
+		"explore_video, explore_body, learn_video, learn_body) VALUES " +
+		"('catcher', 'en', 'catcher_rv', 'catcher_rb', 'catcher_ev', " +
+		" 'catcher_eb', 'catcher_lv', 'catcher_lb')")
+
+	conn.Exec("INSERT INTO book_contents (id, lang, read_video, read_body, " +
+		"explore_video, explore_body, learn_video, learn_body) VALUES " +
+		"('catcher', 'es', 'catcher_es_rv', 'catcher_es_rb', 'catcher_es_ev', " +
+		" 'catcher_es_eb', 'catcher_es_lv', 'catcher_es_lb')")
```
dacruz21, line 58: I think we'll want to move all of this test data back into the test files they are associated with, but I'm not sure how best to accomplish that. This is fine for now, I'll experiment in a separate PR

backend/internal/controllers/middleware/requireAuth.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"context"
+	"net/http"
+	"strings"
+
+	"github.com/TritonSE/words-alive/internal/auth"
+)
+
+// HTTP middleware setting a value on the request context
```
dacruz21, line 11: This comment is so vague, it should definitely mention that it checks the Authorization header on each request against the Authenticator it is given.

backend/internal/controllers/middleware/requireAuth.go
```diff
@@ -0,0 +1,41 @@
...
+func RequireAuth(authenticator auth.Authenticator) func(http.Handler) http.Handler {
+
+	// middleware returned on calling RequireAuth
```
dacruz21, line 14: IMO this comment is unnecessary, we all know what a `return` statement does :) Was there something else you were trying to say here?

backend/internal/controllers/middleware/requireAuth.go
```diff
@@ -0,0 +1,41 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+			// verify the token of the user
+			authHeader := req.Header.Get("Authorization")
+			if authHeader == "" {
+				// kill request
```
dacruz21, line 20: unnecessary I think, `writeResponse` is a pretty common, self-explaining pattern in our codebase

backend/internal/controllers/middleware/requireAuth.go
```diff
@@ -0,0 +1,41 @@
...
+
+			if !strings.HasPrefix(authHeader, "Bearer ") {
+				writeResponse(rw, http.StatusUnauthorized, "User needs to provide Bearer token")
+				return
+			}
+
+			tokenString := strings.TrimPrefix(authHeader, "Bearer ")
+			uid, ok := authenticator.VerifyToken(req.Context(), tokenString)
+			if !ok {
+				// something went wrong
```
dacruz21, line 33: This is also vague, and I think unnecessary

backend/internal/auth/firebase.go
dacruz21: Also, this function does *not* take the entire Authorization header, it just takes the token. Let's rename `authHeader` to `token`

backend/internal/controllers/middleware/utils.go
```diff
@@ -0,0 +1,12 @@
+package middleware
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// Sends data and status code for http
+func writeResponse(rw http.ResponseWriter, statusCode int, body interface{}) {
+	rw.WriteHeader(statusCode)
```
dacruz21, line 10: You need to set the Content-Type header here, just like in `controllers.writeResponse`.

backend/internal/controllers/user.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/TritonSE/words-alive/internal/database"
+	"github.com/TritonSE/words-alive/internal/models"
+)
+
+type UserController struct {
+	// Hold a reference to the database so that all of the functions below can use it
+	// Keeps an authenticator to authenticate token
```
dacruz21, line 14: Not anymore it doesn't, delete this? :)

dacruz21: And actually I think the comment above is unnecessary too, this is a common pattern that's pretty self-documenting

backend/internal/controllers/user.go
```diff
@@ -0,0 +1,59 @@
...
+	Users database.UserDatabase
+}
+
+func (c *UserController) CreateUser(rw http.ResponseWriter, req *http.Request) {
```
dacruz21, line 18: This handler needs a comment though

backend/internal/controllers/user.go
```diff
@@ -0,0 +1,59 @@
...
+		writeResponse(rw, http.StatusForbidden, "Token does not match request body")
+		return
+	}
+
+	_, err := c.Users.FetchUserByEmail(req.Context(), user.Email)
+	if err != nil {
+		// we have created a user with this email before
+		writeResponse(rw, http.StatusBadRequest, "user with email already exists")
+		return
+	}
```
dacruz21, line 42: I don't think this is correct. Right now this will say "user with email already exists" if we can't run the query at all (database is down, SQL syntax error, etc.).

We should keep this `if err != nil` block and throw a 500 in it, but what's more important is checking if the returned user (the first parameter that you've ignored with a `_`) is non-nil, i.e. a user already exists with that email.

backend/internal/controllers/user.go
```diff
@@ -0,0 +1,59 @@
...
+
+	id, err := c.Users.CreateUser(req.Context(), user)
+	if err != nil {
+		writeResponse(rw, http.StatusInternalServerError, "error creating user")
```
dacruz21, line 46: 5xx errors should not have useful messages, can you change this back to `"error"`?

backend/internal/controllers/user.go
```diff
@@ -0,0 +1,59 @@
...
+		fmt.Println(err)
+		return
+	}
+
+	newUser, err := c.Users.FetchUserByID(req.Context(), id)
+	if err != nil {
+		writeResponse(rw, http.StatusInternalServerError, "error fetching user")
```
dacruz21, line 53: Same here

backend/internal/database/users.go
```diff
@@ -0,0 +1,63 @@
+package database
+
+import (
+	"context"
+	// 	"fmt"
```
dacruz21, line 5: Remove this please?

backend/internal/models/user.go
```diff
@@ -0,0 +1,12 @@
+package models
+
+/*
+User model for PostGres backend
+*/
```
dacruz21, line 5: 1. C-style comment syntax isn't necessary for just one line
2. This model isn't specific to Postgres at all, it's intended to be the standard representation of a user across the entire backend.
3. The G in Postgres is lowercase :)

It would be much more useful to explain that this represents a end user of the app (a family, not an admin), and that the ID is the UID of an account in Firebase Auth

backend/internal/testutils/testutils.go
```diff
@@ -3,22 +3,51 @@ package testutils
...
+	req.Header.Add("Authorization", "Bearer "+token)
 
-	err = json.NewDecoder(res.Body).Decode(response)
+	res, err := http.DefaultClient.Do(req)
 	require.NoError(t, err)
+
+	require.Equal(t, expectedStatusCode, res.StatusCode)
+
+	defer res.Body.Close()
+	require.Equal(t, expectedStatusCode, res.StatusCode)
```
dacruz21, line 47: Why are you checking this twice? Also you should defer immediately after require.NoError()

typokign (changes requested):
Tests are really good, thanks for those. This could be better documented though, see comments

backend/internal/controllers/router.go
```diff
@@ -41,5 +45,8 @@ func GetRouter() chi.Router {
 		r.Get("/", bookController.GetBookList)
 	})
 
+	authenticator := auth.MockAuthenticator{}
```
dacruz21, line 48: Let's make GetRouter() take an authenticator as a parameter, and let's define this mock authenticator in main.go, with a big // TODO that we need to set up the Firebase project :)

.gitignore
```diff
@@ -1,2 +1,4 @@
 .idea/
 .DS_Store
+*.env
+commit.txt
```
dacruz21, line 4: not sure how this got in here, remove please? otherwise LGTM!

typokign (approved):
lgtm

backend/internal/auth/mock.go
```diff
@@ -0,0 +1,17 @@
...
+	"strings"
+)
+
+type MockAuthenticator struct{}
+
+// This is an authenticator that takes tokens of the form test-token<user ID>
+func (a MockAuthenticator) VerifyToken(ctx context.Context, token string) (string, bool) {
+
+	if strings.HasPrefix(token, "test-token-") {
+		return strings.TrimPrefix(token, "test-token-"), true
```
k6han, line 14: Incredibly minor but this could be modified so that the structure is
if (problem)
      return false

return true

In the same way that the Firebase VerifyToken is written. Most of our other code also has the structure of checking for problems in if statements, and only returning true at the end

k6han (approved):

k6han:
LGTM. Nothing important needs any changes

PR #18, opened 2021-03-08
Book details screen

allisonbhavsar:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Created individual book view screen
- Added youtube videos and markdown support to tab content
- Added language buttons for tab content

### Testing
- Populated database with dummy data
- Checked that book info rendered properly in tabs (read/learn/explore)
- Tested with multiple languages

### Confirmation of Change 

IMG_8915 6
IMG_8916 2


mobile/src/components/ButtonGroup.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { Colors } from '../styles/Colors';
+import { TextStyles } from '../styles/TextStyles';
+
+// button labels and callback function for passing key of active button to parent
+type ButtonGroupProps = { btn1: string, btn2: string, btn3: string, onBtnChange };
+
+/**
+ * Renders an inline group of three button tabs
+ */
+export const ButtonGroup: React.FC<ButtonGroupProps> = ({ btn1, btn2, btn3, onBtnChange }) => {
```
dacruz21, line 12: I'm not sure how useful this component is. I definitely think we should have components for individual buttons (maybe <SmallButton> and <BigButton>) that match the style guide. I think once you have a component for an individual button it should be easy enough to just define 3 of them in your book details screen and space them manually.

If you decide a ButtonGroup is still helpful though, I think it would be better to take an array of (any number of) buttons, hardcoding it to three buttons is just going to restrict us down the line.

mobile/src/components/ButtonGroup.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const styles = StyleSheet.create({
+  button: {
+    justifyContent: 'center',
+    width: 74,
+    height: 26,
+    borderWidth: 1,
+    borderRadius: 5,
+    borderColor: Colors.orange,
+    shadowColor: 'black',
```
dacruz21, line 60: Let's maybe add all of the colors to Colors.ts, even when they're standard colors like #000000 or #FFFFFF. I would call this `Colors.shadowColor` so we could consistently change the style of all of the shadows across the app if we ever wanted to.

mobile/src/components/LanguageButtons.tsx
```diff
@@ -0,0 +1,93 @@
+import React, { useState } from 'react';
+import { Text, StyleSheet, View, TouchableOpacity } from 'react-native';
+import { Colors } from '../styles/Colors';
+import { TextStyles } from '../styles/TextStyles';
+import { Language, Languages } from '../models/Languages';
+
+// button labels and callback function for passing key of active button to parent
+type LanguageButtonsProps = { langs: Language[], defaultActive: string, onBtnChange };
```
dacruz21, line 8: defaultActive should be a Language, and onBtnChange should have a type

mobile/src/components/LanguageButtons.tsx
```diff
@@ -0,0 +1,93 @@
...
+              onPress={() => {
+                setActiveButton(res);
+                onBtnChange(res);
+              }}
+            >
+              <Text style={[activeButton === res ? styles.buttonTextActive : styles.buttonTextInactive, TextStyles.caption2]}>
+                {Languages[res]}
+              </Text>
+            </TouchableOpacity>
+          </View>
```
dacruz21, line 36: Nice, so you set a max of 3 languages per row and then style the borders appropriately? That sounds good. Let's maybe make 3 a constant though, add a `const LANGS_PER_ROW = 3` outside of the component, just to make the code a little more readable.

Also I would define the styles on line 19 (inside the map(), outside of the return statement) rather than inline it in the JSX, it's a little hard to read.

dacruz21: Also nit: `res` should maybe be called `lang`

mobile/src/navigation/HomeStackNavigator.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { StyleSheet, Image } from 'react-native';
+import { createStackNavigator } from '@react-navigation/stack';
+import { RouteProp } from '@react-navigation/native';
+import { HomeScreen } from '../screens/HomeScreen';
+import { BookScreen } from '../screens/BookScreen';
+import { Language } from '../models/Languages';
+
+type HomeStackParams = {
+  Home: undefined;
+  Book: { id: string, langs: Language[] };
```
dacruz21, line 11: Let's pass the entire Book object into the screen, so we can preload the title, author, and image. You'll still have to call GET /books/id/lang to get the read/explore/learn tabs but at least we can render some content earlier.

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,185 @@
...
+  // get book id and langs from route params
+  const route = useRoute<BookScreenRouteProps>();
+  const { id, langs } = route.params;
+
+  const client = useContext(APIContext);
+  const i18nCtx = useContext(I18nContext);
+
+  // book screen states
+  const [loading, setLoading] = useState(true);
+  const [language, setLanguage] = useState(i18nCtx.locale.substring(0, 2));
```
dacruz21, line 27: So I think the default lang we want here should be (in order of precedence):

1. User's current lang if it exists in book.languages
2. English if it exists in book.languages
3. book.languages[0]

I think you have everything you need to determine this default value before the call to useState, you should be able to do something like:

```ts
const currentLang = i18nCtx.locale.substring(0,2);
const defaultLang = langs.includes(currentLang) ? currentLang : langs.includes('en') ? 'en' : langs[0];
...
const [language, setLanguage] = useState(defaultLang);
```

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const [title, setTitle] = useState('');
+  const [author, setAuthor] = useState('');
+  const [image, setImage] = useState('');
+  const [tabBody, setTabBody] = useState('');
+  const [tabVideo, setTabVideo] = useState('');
+  const [activeButton, setActiveButton] = useState('');
+  const [tabContent, setTabContent] = useState(null);
```
dacruz21, line 34: Any reason you split these all out into separate state variables instead of having a single `const [book, setBook] = useState()`? I think that would be a lot cleaner and (maybe?) more performant. React can combine state updates but it's not guaranteed, so we might get a lot of re-renders with separate vars.

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const markdownStyles = {
+    heading: { fontFamily: 'Gotham-Light' },
+    paragraph: { fontFamily: 'Gotham-Light' },
+    strong: { fontFamily: 'Gotham-Bold' },
+    listItemNumber: { fontFamily: 'Gotham-Light' },
+    listItemOrderedContent: { fontFamily: 'Gotham-Light' },
+    listItemUnorderedContent: { fontFamily: 'Gotham-Light' },
+    tableHeaderCellContent: { fontFamily: 'Gotham-Light' },
+    em: { fontFamily: 'Gotham-Italic' },
+  };
```
dacruz21, line 45: Can we move these to TextStyles.ts or does that cause any issues?

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,185 @@
...
+      if (langs.indexOf(language) > -1) {
+        lang = language;
+      } else if (langs.indexOf('en') > -1) {
+        lang = 'en';
+        setLanguage(lang);
+      } else {
+        const [first] = langs;
+        lang = first;
+        setLanguage(lang);
+      }
```
dacruz21, line 62: Ah here it is, yeah I think you can just determine this before the call to useState().

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,185 @@
...
+        setActiveButton('btn-1');
+        setTabContent({
+          'btn-1': res.read,
+          'btn-2': res.explore,
+          'btn-3': res.learn,
+        });
+        setLoading(false);
+      }).catch((err) => {
+        console.log(err);
+      });
```
dacruz21, line 77: Can you async/await this please? useEffect can't take an async function unfortunately so you'll have to wrap this in an async IIFE

```ts
useEffect(() => {
  (async () => {
    // do stuff
  })();
}, [language]);
```

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,185 @@
...
+          const matches = tabContent[activeButton].video.match(regEx);
+          if (matches) {
+            setTabVideo(matches[1]);
+          }
+        }
+        setTabBody(tabContent[activeButton].body);
+      }
+    },
+    [activeButton, tabContent],
+  );
```
dacruz21, line 98: I think you should have a state var for the selected tab of type `'read' | 'explore' | 'learn'`, then you won't need an effect. You can just pass `book[selectedTab]` into the youtube and markdown renderers

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  return (
+    loading ? <View style={styles.loadingCircle}><LoadingCircle /></View> :
```
dacruz21, line 101: nit: it might be better to just

```ts
if (loading) {
  return <View style={styles.loadingCircle}><LoadingCircle /></View>;
}

return (
  // regular content
)
```

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,185 @@
...
+      (
+        <ScrollView showsVerticalScrollIndicator={false}>
```
dacruz21, line 103: Any reason to turn the scrollbar off?

typokign (changes requested):
It looks gorgeous! :heart_eyes: 

I think the screen is a bit overly complex though, you should be able to get rid of a lot of the effects and combine some state variables.

mobile/src/screens/BookScreen.tsx
dacruz21: Oh wait, not quite. Since we want to render the title/author/image while content is loading (see above) we don't want to just hide the entire screen if it's loading.

Leave this as one return statement but wrap the video and markdown components in this `loading ? : ` guard

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,185 @@
...
+import { TextStyles } from '../styles/TextStyles';
+import { ButtonGroup } from '../components/ButtonGroup';
+import { LanguageButtons } from '../components/LanguageButtons';
+
+/**
+ * Individual book view displaying book details
+ */
+export const BookScreen: React.FC = () => {
+  // get book id and langs from route params
+  const route = useRoute<BookScreenRouteProps>();
```
dacruz21, line 19: You can just add the route object to the arguments of BookScreen, no need for a hook here: https://reactnavigation.org/docs/params

```ts
export const BookScreen: React.FC = ({ route }) => {

}
```

mobile/src/navigation/HomeStackNavigator.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { HomeScreen } from '../screens/HomeScreen';
+import { BookScreen } from '../screens/BookScreen';
+import { Book } from '../models/Book';
+
+type HomeStackParams = {
+  Home: undefined;
+  Book: { book: Book };
+};
+
+export type BookScreenRouteProps = RouteProp<HomeStackParams, 'Book'>;
```
dacruz21, line 14: Damn the TS definitions are annoying for React Navigation, but I think I found a good way to type the BookScreen props. You should be able to get rid of the BookScreenRouteProps type here, and export HomeStackParams, and then...

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { I18nContext } from '../context/I18nContext';
+import { TextStyles } from '../styles/TextStyles';
+import { ButtonGroup } from '../components/ButtonGroup';
+import { LanguageButtons } from '../components/LanguageButtons';
+import { Colors } from '../styles/Colors';
+
+/**
+ * Individual book view displaying book details
+ */
+export const BookScreen: React.FC = () => {
```
dacruz21, line 18: Put this here:

```ts
type BookScreenProps = StackScreenProps<HomeStackParams, 'Book'>;

/**
 * Individual book view displaying book details
 */
export const BookScreen: React.FC<BookScreenProps> = ({ route }) => {
  // get book id and langs from route params
  const { book } = route.params;
  const langs = book.languages;
```

mobile/src/screens/BookScreen.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const locale = i18nCtx.locale.substring(0, 2);
+  const defaultLang = langs.includes(locale) ? locale : langs.includes('en') ? 'en' : langs[0];
+
+  const videoIdRegEx = '^(?:https?:)?//[^/]*(?:youtube(?:-nocookie)?.com|youtu.be).*[=/]([-\\w]{11})(?:\\?|=|&|$)';
+
+  // book screen states
+  const [loading, setLoading] = useState(true);
+  const [bookDetails, setBookDetails] = useState(null);
+  const [language, setLanguage] = useState(defaultLang);
+  const [activeButton, setActiveButton] = useState('read');
```
dacruz21, line 36: Can you type these for a little extra safety? There are a lots of `any`s being thrown around here and I'm surprised TS hasn't been reporting some of them as errors...

You don't need to type `loading` since that's obviously a bool (and TS knows that), but the rest of these need to be more explicit:

```ts
  const [bookDetails, setBookDetails] = useState<BookDetails>(null);
  const [language, setLanguage] = useState<Language>(defaultLang);
  const [activeButton, setActiveButton] = useState<Tab>('read');
```

For the Tab type, you can define that somewhere at the top of this file as `type Tab = 'read' | 'explore' | 'learn'`

typokign (changes requested):
The screen component is looking much cleaner now, thank you! I think the last little things are just type definitions.

dacruz21:
Just 2 more things I found testing the Markdown renderer:

1. Let's enable links by setting the onLinkPress prop: https://www.npmjs.com/package/react-native-markdown-view?activeTab=readme#onlinkpress. You can use the WebBrowser API to open links within the app: https://docs.expo.io/versions/latest/sdk/webbrowser/. This is better than the Linking API for this purpose since it doesn't switch apps into your web browser, and it lets you press the back button to return to the app.
2. When there's no video, the markdown content is centered, which I don't think we want. You can add `width: '100%'` to your tabContentContainer style and change the horizontal margins to paddings to fix that

typokign (dismissed):
Great work!

dacruz21:
Sorry to send this back to requested changes, but I'm adding books to staging right now and I found just a few stylistic things:

1. The Gotham-Italic font doesn't look right for inline italics, the text is much bolder than the default font. Can you replace the fontFamily override for mdEm with a fontStyle: 'italic'?
2. You can't really control image size or alignment in Markdown, so I think a good default would be to have the image take up the entire width of the screen. WA uses images instead of YouTube videos for some tabs to give them some color. Can you add `imageWrapper: { width: '100%' }` to the markdownStyles object?
3. The spacing around bullet points and list numbers are huge on my phone, can you set their width to 16 (default is 32)? In markdownStyles:

```ts
listItemNumber: {
    minWidth: 16,
  },
  listItemBullet: {
    minWidth: 16,
  },
  ```

**Review by typokign — DISMISSED** [review]
Great work!

File: mobile/src/navigation/MainTabNavigator.tsx [hunk 18]
```diff
@@ -0,0 +1,82 @@
...
+
+const styles = StyleSheet.create({
+
+  navbarContainer: {
+    height: 60,
+    padding: 4.5,
+    paddingLeft: 47,
+    paddingRight: 47,
+    backgroundColor: '#FFFFFF',
+    shadowColor: '#000000',
```
aksharans, line 53: Colors.shadowColor

aksharans (changes requested):
Just a couple syntax stuff in the stylesheet for MainTabNavigator.
I've been using Pressable for the buttons, not sure if there was any performance/support difference with TouchableOpacity 
( @dacruz21 ).
But other than that, looks pretty good to me.

aksharans (approved):
lgtm

typokign (approved):

PR #19, opened 2021-03-10
Chat Screen up to date with Hi-Fi mockups

RohithKasar:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Added Chat Contact screen page

### Testing
- Passes checks

### Confirmation of Change 
Chat Screen Page

Image from iOS


mobile/src/screens/ChatScreen.tsx
```diff
@@ -1,21 +1,31 @@
...
 export const ChatScreen: React.FC = () => {
   return (
     <View style={styles.container}>
-      <Text>Chat</Text>
+      <Text style={TextStyles.h3}>  {`\tLet's Talk\n`}</Text>
+      <Text> <Text style={TextStyles.h3}> {`\t`}Tel: </Text><Text>858.274.9673{`\n`}</Text> </Text>
+      <Text> <Text style={TextStyles.h3}> {`\t`}Fax: </Text><Text>858.274.9673{`\n`}</Text> </Text>
+      <Text> <Text style={TextStyles.h3}> {`\t`}Mon-Fri: </Text><Text>8am-4pm{`\n`}</Text> </Text>
+      <Text> <Text style={TextStyles.h3}> {`\t`}Email Us: </Text><Text>info@wordsalive.gov{`\n`}</Text> </Text>
+      <Text> <Text style={TextStyles.h3}> {`\t`}Address: </Text><Text>{`5111 Santa Fe Street Suite\n\t219 San Diego, California, 92081, \n\tUnited States`}{`\n`}</Text> </Text>
```
dacruz21, line 16: Why are these double-wrapped in `<Text>` components?

mobile/src/screens/ChatScreen.tsx
```diff
@@ -1,21 +1,31 @@
...
 
 const styles = StyleSheet.create({
   container: {
     flex: 1,
-    alignItems: 'center',
+    alignItems: 'flex-start',
     justifyContent: 'center',
+    backgroundColor: 'orange',
+    width: '400pt',
+    height: '400pt',
```
dacruz21, line 29: This is crashing my phone, I don't think "pt" is a unit in RN. What is this supposed to do?

mobile/src/screens/ChatScreen.tsx
```diff
@@ -1,21 +1,31 @@
...
   );
 };
 
 const styles = StyleSheet.create({
   container: {
     flex: 1,
-    alignItems: 'center',
+    alignItems: 'flex-start',
     justifyContent: 'center',
+    backgroundColor: 'orange',
```
dacruz21, line 27: Colors.orange for consistency please

mobile/src/screens/ChatScreen.tsx
dacruz21: Also as discussed, you should use CSS for styling, not tabs and line breaks.

dacruz21: Lastly, this is only a temporary screen but we should still make all of the text internationalized, can you add these strings to i18n/?

typokign (changes requested):

mobile/src/screens/ChatScreen.tsx
dacruz21: Also you're missing text styles on some of these text tags (the parts after the colon) so the font doesn't match. Can you define a new style with the same size but with the light font variant?

dberror.patch
```diff
@@ -0,0 +1,86 @@
+diff --git a/backend/internal/controllers/middleware/requireAuth.go b/backend/internal/controllers/middleware/requireAuth.go
```
dacruz21, line 1: Remove this file please

mobile/assets/i18n/en.json
```diff
@@ -1,6 +1,16 @@
...
     "newBooks": "New Books for You",
     "allBooks": "All Books",
+    "letsTalk": "Let's Talk",
+    "tel": "Tel: ",
+    "num": "858.274.9673",
+    "fax": "Fax: ",
+    "days": "Mon-Fri: ",
+    "hours": "8am-4pm",
+    "emailUs": "Email Us: ",
+    "emailAddress": "info@wordsalive.gov",
```
dacruz21, line 11: I didn't know TSE was taking on government contracts ;)

mobile/assets/i18n/en.json
```diff
@@ -1,6 +1,16 @@
...
+    "addressTitle": "Address: ",
+    "actualAddress": "Santa Fe Street Suite 219 San Diego, California, 92081, United States",
```
dacruz21, line 13: I feel like the names and values of these translations could be refactored a little more clearly.

First of all, stuff like punctuation doesn't need to be translated, so I think we can move the colons here outside of i18n.

I also don't think the address or email address will need to be translated - they're really only valid in English. Phone number arguably _maybe_ could be translated so that we could include the area code internationally, but langauge != location, so maybe we should just add the area code to the number in all languages. Lmk your thoughts here.

Lastly, i18n translation keys are usually just the English value in camel case, which makes the code easier to read. I think if you remove the num, emailAddress, and actualAddress keys that should be a pretty easy change.

mobile/assets/i18n/en.json
```diff
@@ -1,6 +1,16 @@
 {
     "newBooks": "New Books for You",
     "allBooks": "All Books",
+    "letsTalk": "Let's Talk",
+    "tel": "Tel: ",
+    "num": "858.274.9673",
+    "fax": "Fax: ",
+    "days": "Mon-Fri: ",
+    "hours": "8am-4pm",
```
dacruz21, line 9: Oh, I think we might be able to do this automatically! https://github.com/fnando/i18n-js#date-formatting

mobile/src/screens/ChatScreen.tsx
```diff
@@ -1,21 +1,51 @@
...
+    marginRight: '20%',
+    backgroundColor: Colors.orange,
+    width: '100%',
+    height: '100%',
+  },
+  title: {
+    marginLeft: '6%',
+    padding: '2%',
+    width: '88%',
   },
```
dacruz21, line 50: I don't understand some of the values down here - why the 20% right margin? Why the padding and width on titles?

mobile/assets/i18n/en.json
dacruz21, line 13: You're missing the 5111 here

mobile/src/screens/ChatScreen.tsx
```diff
@@ -1,21 +1,51 @@
...
 /**
  * Left tab on navbar for chatting with volunteers
  */
 export const ChatScreen: React.FC = () => {
+  const i18nCtx = useContext(I18nContext);
   return (
     <View style={styles.container}>
-      <Text>Chat</Text>
+
+      <Text style={[TextStyles.h2, styles.title]}> {i18nCtx.t('letsTalk')}</Text>
```
dacruz21, line 15: In the mockup, Let's Talk definitely looks larger than the rest of the text - let's bump this up to h1

mobile/src/screens/ChatScreen.tsx
dacruz21: Ok I played with this locally and I see what some of these are for, but I think a lot of it is unnecessary. I would personally:

- Drop the styles.title, I don't think you need it
- Add paddingLeft: 24, and paddingRight:24 to the container

mobile/src/screens/ChatScreen.tsx
```diff
@@ -1,21 +1,51 @@
...
+      </Text>  
+      <Text style={[TextStyles.h2, styles.title]}> 
+        {i18nCtx.t('days')} <Text style={TextStyles.h3}>{i18nCtx.t('hours')}</Text>
+      </Text>
+      <Text style={[TextStyles.h2, styles.title]}> 
+        {i18nCtx.t('emailUs')} <Text style={TextStyles.h3}>{i18nCtx.t('emailAddress')}</Text>
+      </Text>
+      <Text style={[TextStyles.h2, styles.title]}> 
+        {i18nCtx.t('addressTitle')} <Text style={TextStyles.h3}>{i18nCtx.t('actualAddress')}</Text>
+      </Text>
```
dacruz21, line 31: So looks like you're doing a constant padding of 2% on all lines here, which doesn't match the mock up. I think we should stay true to the mockup here, especially since the hours (Mon-Fri 8am-4pm) are supposed to be the hours for the phone hotline, so I think we should keep them clustered.

Now in terms of actually implementing this, I tried adding React fragments and `<View/>`s everywhere I wanted a double-space, but the only thing that seemed to work was an empty `<Text/>`. That's probably fine (maybe leave a comment saying it's for padding though?) but if you have any other "cleaner" ideas let me know.

typokign (changes requested):

dacruz21:
Lastly let's add the "Chat is coming soon, until then:" flavor text we talked about so the "chat" tab leading to a phone number doesn't feel like a complete bait and switch lol.

dberror.patch
dacruz21: Still here

mobile/src/screens/ChatScreen.tsx
```diff
@@ -1,21 +1,71 @@
...
 /**
  * Left tab on navbar for chatting with volunteers
  */
 export const ChatScreen: React.FC = () => {
+  const i18nCtx = useContext(I18nContext);
   return (
     <View style={styles.container}>
-      <Text>Chat</Text>
+      <Text style={TextStyles.heading1}>Live chat is coming soon</Text>
+      <Text style={TextStyles.heading2}>Until then...</Text>
```
dacruz21, line 15: i18n this please

typokign (changes requested):

mobile/src/styles/TextStyles.ts
```diff
@@ -18,14 +18,19 @@ export const TextStyles = StyleSheet.create({
...
   },
   body2: {
     fontSize: 12,
     fontFamily: 'Gotham-Light',
   },
+  body3: {
+    fontSize: 18,
+    fontFamily: 'Gotham-Light',
+    fontWeight: 'normal',
+  },
```
dacruz21, line 33: Definitely seems weird to have body3 be *larger* than body1 and 2, can you move this into ChatScreen? I don't think we're going to use this anywhere else

typokign (approved):
approved

Anshul-Birla (approved):
Everything looks good

PR #20, opened 2021-03-11
Home page All Books with pagination and filters

aksharans:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Added PaginationBookList (replacing ColumnBookList) to display "paginated" books 
- Added BookSearchFilter to filter books by search term
- Added Language Filter to filter books by language
- Added AllBooksWithFilters component to encapsulate the above components and display on the Home Screen

### Testing
- tested pagination and search/language filters on phone using test book data and various user input cases

### Confirmation of Change 

Deleted the test data for the PR, but to see the change: 
add the code from the commit "delete test data" (the array of books and dispatch call) to the BookContext.tsx file, and run npm start.


dacruz21:
Nice! Visual stuff I noticed:

1. The hamburger icon is too big, can you get it closer to the mockup please?
2. IMO "Search for a book by title or author" is pretty verbose and might even get cut off on some screens, maybe reword to "Search for a title or author"?
3. The checkboxes don't work on my phone, but they do on web... I'll look at the code later and try to guess why that is.
4. Even on web, tapping the labels for the checkboxes doesn't toggle the checkbox. Since the checkboxes are pretty small touch targets, I think you should wrap each row in an onPress, not just the checkboxes.
5. Similar to the touch target issue, I think the font size on the language picker is too small to easily press on my phone.
6. Tapping outside of the language picker doesn't close it. Tbh I'm not sure what the cleanest way to fix this is, but it's probably fine to just wrap the entire screen in an onPress (animations disabled) that closes the menu.

mobile/src/components/PaginationBookList.tsx
```diff
@@ -0,0 +1,91 @@
...
+                    <View style={styles.emptyBook} />
+                  </View>
+                )
+            ))}
+
+          </View>
+
+          <View style={styles.text}>
+            <Text style={TextStyles.h3}> {'<'} {index + 1} {'>'} </Text>
+            <Text style={{ ...TextStyles.h1, fontSize: 10 }}>of {booksChunked.length}</Text>
```
dacruz21, line 57: Let's remove this for now since it's not on the mockup, but I will ask design and WA for feedback when I show it to them.

mobile/src/screens/HomeScreen.tsx
```diff
@@ -1,26 +1,28 @@
...
 export const HomeScreen: React.FC = () => {
   // get books from backend
   const booksCtx = useContext(BookContext);
   useEffect(booksCtx.fetchBooks, []);
   const newBooks = booksCtx.books
-    .sort((a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime())
+    .sort((a, b) => new Date(b.created_at).getTime() - new Date(a.created_at).getTime())
     .slice(0, 5);
   const allBooks = booksCtx.books;
+  const { loading } = booksCtx;
```
dacruz21, line 25: Might as well combine these two into one line

mobile/src/components/AllBooksWithFilters.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+// object containing the array of language objects for use as a state
+const langState = {
+  arr: langArray,
+};
+
+/**
+ *  Renders the All Books section with a search filter and displays the paginated books.
+ */
+export const AllBooksWithFilters: React.FC<AllBooksWithFiltersProps> = ({ allBooks, loading }) => {
```
dacruz21, line 38: Yeah I think we can remove this component entirely. All of the dropdown and filtering logic can go into the hook, and the "No results" text can just go into the home screen

mobile/src/components/AllBooksWithFilters.tsx
```diff
@@ -0,0 +1,206 @@
...
+// for use in the filter with default of "not active"
+const langArray = [];
+langCodes.forEach(el => {
+  langArray.push({ lang: el, isActive: false });
+});
+
+// object containing the array of language objects for use as a state
+const langState = {
+  arr: langArray,
+};
```
dacruz21, line 33: This is the problem with the checkboxes, these need to be proper state variables inside components. I would add a Set of selected languages to the filter hook, like:

```ts
const [selectedLangs, setSelectedLangs] = useState(new Set<Language>());

// on select lang:
setSelectedLangs(langs => new Set(langs).add('en'));

// on deselect
setSelectedLangs(langs => new Set(langs).delete('en'));
```

typokign (changes requested):
I think I found the source of the checkbox bug, see below. Sorry if I didn't explain this when I showed you an example custom hook, but you should put all of the book filtering logic and components (text search and language filter) into the hook. Otherwise you still need to write a lot of logic in the AllBooksWithFilters component and you're not getting much value from it.

aksharans:
> 5. Similar to the touch target issue, I think the font size on the language picker is too small to easily press on my phone.
Should I make the size of the enclosing dropdown dynamic based on the max language length (if more languages are added) or just a static value?

mobile/src/components/AllBooksWithFilters.tsx
aksharans: For the separate component, if I use the hook directly in the HomeScreen, every time a state changes, the screen "jumps" to the top, but moving it to the AllBooksWithFilters makes it so that the state change doesn't move the position of the screen. Is there a way of maintaining the screen position or should I just keep the component?

dacruz21: Huh, that shouldn't happen. Maybe there's a moment during the change in state where the list of books is empty, and the view collapses to height=0, shifting the screen? You can get around that by just giving the view a minimum height when there are no books. Just my guess, if that doesn't help try removing this component anyway and I can help debug.

dacruz21:
I don't think you should need to calculate the size of the View itself, you should be able to do some CSS magic to get it to shrinkwrap to the contained elements: https://stackoverflow.com/questions/36979550/how-to-make-html-element-wrap-its-content-and-shrink-to-fit-afterwards

Hopefully this works in RN, I don't think I've ever tested it.

typokign (changes requested):
Thanks for fixing conflicts, but there are a few things we need to fix now:

- You started this branch before the book details screen was done, so your PaginatedBookList doesn't have the navigation logic. Take a look at the Pressables in ColumnBookList and add them to your PaginatedBookList
- Something about this rebase broke your language selector, the checkboxes aren't checking when I tap them.
- Can we make the hamburger button a TouchableOpacity to get that nice touch feedback?
- Can we put a little bit of left padding on the "Search for a title or author" text?

mobile/src/components/BookFilter.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+          <View style={styles.nameBoxContainer}>
+
+            <Text style={{ ...TextStyles.caption4, alignSelf: 'center' }}>{Languages[lang]}</Text>
+
+            <View style={styles.box}>
+              {langs.has(lang) ?
+                <Image style={styles.boxChecked} source={require('../../assets/images/check-square-solid.png')}/>
+                :
+                null}
```
dacruz21, line 63: This is fine, just a tip that this can be reduced to `langs.has(lang) && <Image ...>`

mobile/src/context/BookContext.tsx
```diff
@@ -20,13 +20,208 @@ export const BookProvider: React.FC = ({ children }) => {
...
 
-  // loads books. If error, logs erorr and returns a loading indicator of true
+  // loads books. If error, logs error and returns a loading indicator of true
   function fetchBooks(): void {
     dispatch({ type: 'API_CALL_STARTED' });
     client.getBooks().then((res) => {
       dispatch({ type: 'API_RETURNED', payload: res });
     }).catch((err) => {
       console.log(err);
+      const books: Book[] = [
```
dacruz21, line 30: Ready to remove this? Also is there something wrong with `npm run staging`? I would think that would be an easier way to get some books without running the backend

mobile/src/context/BookContext.tsx
```diff
@@ -44,8 +239,8 @@ export const BookProvider: React.FC = ({ children }) => {
 };
 
 type BookAction
-  = { type: 'API_CALL_STARTED' }
-  | { type: 'API_RETURNED', payload: Book[] };
+ = { type: 'API_CALL_STARTED' }
+ | { type: 'API_RETURNED', payload: Book[] };
```
dacruz21, line 243: Revert/reindent please

mobile/src/screens/HomeScreen.tsx
```diff
@@ -1,85 +1,147 @@
...
 import { BookContext } from '../context/BookContext';
 import { TextStyles } from '../styles/TextStyles';
 import { Colors } from '../styles/Colors';
 import { I18nContext } from '../context/I18nContext';
-import { LoadingCircle } from '../components/LoadingCircle';
+import { Language } from '../models/Languages';
+import { Book } from '../models/Book';
+
+// how many books to display per page (in the Paginated Booklist)
+const booksPerPage = 9;
```
dacruz21, line 18: Let's just make this a constant in the PaginatedBookList file, rather than a prop. Also, when you move it, CONST_CASE it please

mobile/src/screens/HomeScreen.tsx
```diff
@@ -1,85 +1,147 @@
...
+
+const { width } = Dimensions.get('window');
 
 /**
- * Renders the homescreen for the app. Currently displays heading, new books, all books.
- */
+* Renders the homescreen for the app. Currently displays heading, new books, all books.
+*/
```
dacruz21, line 24: Did you override the indentation settings in your editor or something? Do you mind cleaning up all the small indentation changes?

mobile/src/components/PaginatedBookList.tsx
```diff
@@ -0,0 +1,102 @@
...
+const { width } = Dimensions.get('window');
+
+type PaginatedBookListProps = { books: Book[], booksPerPage: number };
+
+/**
+* Renders a "paginated" list of books where books are grouped into grids and displayed on multiple pages.
+*/
+export const PaginatedBookList: React.FC<PaginatedBookListProps> = ({ books, booksPerPage }) => {
+  // "chunks" the books into groups of size <booksPerPage> and fills any empty spots
+  const booksChunked: Book[][] = [];
```
dacruz21, line 19: Maybe rename this to `pages` or `bookPages`?

typokign (changes requested):
Nice! Works really well and is very clean. Just clean up the mocked books and we should be g2g

mobile/src/screens/HomeScreen.tsx
aksharans: I think it was because the virtualizedView, which was unindendented was replaced with the scrollView, which was indented, so all its children got another space of indentation.

mobile/src/context/BookContext.tsx
aksharans: 👍 , Nothing wrong with staging, I was just trying to test different inputs and make sure the expected books were showing up.

mobile/src/screens/HomeScreen.tsx
aksharans: I had the booksPerPage in HomeScreen bc the loading/no results view was also there, so I wanted to make sure the heights were the same with/without books so the screen doesn't get cut off.

typokign (approved):
lgtm

rohanputhukudy:
LGTM

rohanputhukudy (approved):

PR #23, opened 2021-03-28
Get/Update Profile API

k6han:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Added GET and PATCH requests for the users database

### Testing
- Added tests to the user_test.go file in the controllers directory

### Confirmation of Change 
- Should be able to use GET requests to see users (if authenticated) and use PATCH to update user info


backend/internal/controllers/user.go
```diff
@@ -65,3 +67,95 @@ func (c *UserController) CreateUser(rw http.ResponseWriter, req *http.Request) {
...
+}
+
+/*
+ * Handles PATCH request, requires {id}
+ * Update fields of user info
+ */
+func (c *UserController) UpdateUser(rw http.ResponseWriter, req *http.Request) {
+	var userID string = chi.URLParam(req, "id")
+
+	var user models.User
```
dacruz21, line 116: Hmm, I think we should define a new struct for this route. Re-using models.User here has two consequences:

1. It allows the request to specify a new ID or email, which isn't allowed, so you have to prevent this manually as you've done before
2. It doesn't allow you to specify null values, making partial updates (which is something PATCH should allow) impossible.

Instead, let's make an UpdateUser struct that includes only the fields you're allowed to update, and gives each field a pointer type (so they can be nil in partial updates).

backend/internal/controllers/user_test.go
```diff
@@ -39,3 +41,53 @@ func TestCreateDuplicateUser(t *testing.T) {
...
+}
+
+func TestGetUserUnAuth(t *testing.T) {
+	var response string
+	testutils.MakeAuthenticatedRequest("GET", ts.URL+"/users/user1", "", http.StatusForbidden, &response, "test-token-user2", t)
+}
+
+func TestUpdateUser(t *testing.T) {
+	// valid token test
+	body := `{"id": "user1", "name": "roberto", "email": "test4@test.com", "in_san_diego": false}`
```
dacruz21, line 68: Once you do the above let's check that it actually works by writing two tests that do partial updates. Drop the id and email fields from this request body, and replace it with two tests with bodies: `{"name":"new name"}` and `{"in_san_diego": false}`

typokign (changes requested):
Nice, love the coalesce() function :) But it's not going to work if the JSON schema doesn't allow nulls, see below:

typokign (dismissed):
approved

backend/internal/controllers/user.go
```diff
@@ -65,3 +67,97 @@ func (c *UserController) CreateUser(rw http.ResponseWriter, req *http.Request) {
...
+		writeResponse(rw, http.StatusNotFound, "user not found")
+		return
+	}
+	/*
+		// Illegal attempt to change email
+		if user.Email != "" && user.Email != currUser.Email {
+			writeResponse(rw, http.StatusBadRequest, "cannot change email")
+			return
+		}
+	*/
```
RohithKasar, line 153: Everything looks good, maybe delete this comment? Or are we saving it for future implementation?

RohithKasar (dismissed):
This is really nice, very readable and understandable code. There was commented out code in one file that you can keep if you want but other than that it looks good!

backend/internal/controllers/user.go
dacruz21: Oops sorry, missed this. Yeah @k6han please delete :)

typokign (dismissed):

RohithKasar (dismissed):

typokign (approved):

RohithKasar (approved):

PR #24, opened 2021-03-28
Continuous deployment pipeline

typokign:
### Administrative Info
Monday Board ID: 853642866

### Changes
Added continuous deployment pipeline to Heroku. Commits to main get auto-deployed to staging, tagged commits get auto-deployed to production.

### DIFF
--- a/.github/workflows/words-alive.yml
+++ b/.github/workflows/words-alive.yml
@@ -3,6 +3,7 @@ name: words-alive
 on:
   push:
     branches: [ main ]
+    tags: [ v** ]
   pull_request:
     branches: [ main ]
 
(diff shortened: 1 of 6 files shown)


Anshul-Birla (dismissed):
I like it, but how do the tagged commits work? do we have to commit a certain way now? And should we split this into two different files instead of just one so that the CD and CI pipelines are different?

dacruz21:
> I like it, but how do the tagged commits work? do we have to commit a certain way now?

No, git tags are different from git commits. When we want to release a new version we'll just run `git tag v1.2.3` and `git push origin --tags`.

> And should we split this into two different files instead of just one so that the CD and CI pipelines are different?

I really wanted to but it looks like there's no easy way to specify dependencies between different workflows (.yml files), they just all get run in parallel. Keeping it in one file also makes it easier to visualize with this cute graph view:

image

Anshul-Birla:
> No, git tags are different from git commits. When we want to release a new version we'll just run `git tag v1.2.3` and `git push origin --tags`.

Do you want to just automate it with something like this https://github.com/semantic-release/semantic-release. Used it for my 110 project. Worked pretty well. Just need to commit a specific way and it takes care of versioning and stuff,

dacruz21:
Oh nice, I think I've used something like this before. Yeah this looks nice for generating changelogs but I don't think it helps with actually deploying to something like Heroku, just releasing npm packages.

I'll put this down as a maybe for later but at the same time our commit message hygiene is really bad and I don't see people actually following this format lul.

k6han (dismissed):
LGTM

k6han (dismissed):
LGTM

dacruz21:
@allisonbhavsar CORS fix is in this PR, please review and merge so this gets pushed out to staging for you :)

allisonbhavsar (approved):
lgtm

Anshul-Birla (approved):
Looks good

PR #27, opened 2021-04-18
Added image API 

Anshul-Birla:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Adde POST and GET route for images 
- Added blurhash strings for id

### Testing
- Tested post with jpg and png 
- Tested get for all images 
- Stored blurhash images to disk to see if they were blurred correctly 

### Confirmation of Change 
-  Ran curl command: `curl --data-binary "@./airplane.jpg" -H "Content-Type: image/jpeg" -v -XPOST localhost:8080/image` with different images. Visit the url returned to see the image in browser


backend/migrations/20210408134423_add_image_table.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE image (
```
dacruz21, line 3: Let's call this `images` to keep with conventions

backend/migrations/20210408134423_add_image_table.sql
```diff
@@ -0,0 +1,13 @@
...
+    id text PRIMARY KEY,
+    img bytea NOT NULL,
+    mime_type text not NULL
+);
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+DROP TABLE image
```
dacruz21, line 12: This needs a semicolon I think

backend/internal/database/img.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+type ImgDatabase struct {
+	Conn *pgxpool.Pool
+}
+
+/*
+ * Gets an image from the image table given the id. If image exists, returns the binary data of the image
+ */
+func (db *ImgDatabase) GetImage(ctx context.Context, id string) (*[]byte, *string, error) {
```
dacruz21, line 23: Any reason this returns pointers? If it's because of the potential for errors, it's usually more idiomatic to just return `nil, "", err` - by convention the first two values are always ignored if err != nil.

backend/internal/database/img.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	return &img, &ctype, nil
+
+}
+
+/*
+ * Puts an image into the image table and returns id (generated using blurhash) and
+ * a boolean specifying if an image was entered into the db. If the image
+ * already exists, returns hash of pre-existing image
```
dacruz21, line 41: I think it would be better for the controller to do a `GetImage` to see if the image already exists, rather than sending a few MB (potentially) to the database only to find out the file already exists.

backend/internal/database/img.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return nil, false, errors.Wrap(err, "error on decoding image")
+	}
+
+	// encode using blurhash
+	hash, err := blurhash.Encode(4, 4, img)
+
+	if err != nil {
+		return nil, false, errors.Wrap(err, "error on encoding with blurhash")
+	}
```
dacruz21, line 59: Can we do all the image processing in the controller, instead of in the database-specific logic? This feels like a little bit of mixing of concerns.

backend/internal/controllers/img.go
```diff
@@ -0,0 +1,74 @@
...
+// gets an image from the database (/image/{id})
+func (c *ImgController) GetImage(rw http.ResponseWriter, req *http.Request) {
+	var id string = chi.URLParam(req, "id")
+
+	img, ctype, err := c.Image.GetImage(req.Context(), id)
+
+	if err != nil {
+		writeResponse(rw, http.StatusNotFound, "Image with requested id not found")
+		return
+	}
```
dacruz21, line 31: This is potentially misleading - if the database is down it will return 404s on images that exist, instead of a more-appropriate 500.

Instead can you maybe return `img=nil, ctype="", err=nil` for the case where the image doesn't exist, and reserve `err!=nil` for database errors?

backend/internal/controllers/img.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+type ImgController struct {
+	Image database.ImgDatabase
+}
+
+var allowedTypes = map[string]bool{
+	"image/png":  true,
+	"image/jpeg": true,
+}
```
dacruz21, line 20: how dare you, this is so shamefully wasteful, you could save TWO WHOLE BYTES by converting this to a `map[string]struct{}`, and using `if _, found := allowedTypes[key] {`

kidding, this is fine, you can leave it if you want, just a nice pro tip in case you ever need a much larger set with millions of entries

typokign (changes requested):
Really nice! Mainly just some suggestions on go idioms below.

One more thing though, let's also add a maximum file size - Cloudflare will cap us at 100MB per request but even that's way bigger than we'll ever actually need. I'm thinking maybe something like 5MB? That's just a guess, if you want to do some research and figure out a reasonable limit for around 1080p images feel free to adjust.

Also you have a `mobile/.package.json.swp` file that found its way into your branch.

typokign (dismissed):
LGTM, can you ping someone else for review?

typokign (dismissed):

backend/internal/database/img.go
```diff
@@ -0,0 +1,57 @@
...
+		// when image cannot be found in the database
+		if err.Error() == "no rows in result set" {
+			return nil, "", nil
+		}
+
+		return nil, "", errors.Wrap(err, "error in SELECT")
+	}
+
+	return &img, "", nil
+
```
k6han, line 40: What is the variable ctype here? Did you mean to return it?

k6han: If ctype isn't meant to be returned, I think we can remove the string from the return tuple

k6han (changes requested):
Seems to be either an unused variable or an extra return value in database/img.go:GetImage
Aside from that, everything else looks fine

backend/internal/database/img.go
Anshul-Birla: good catch il fix it asap

typokign (approved):

k6han (commented):
LGTM

k6han (approved):
LGTM

PR #30, opened 2021-04-30
Add pre-commit lint hook

justinyaodu:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Add Git pre-commit script which runs ESLint
- Add Husky to automatically manage Git hook installation

### Testing
- Made a commit and confirmed that ESLint ran
- Created a lint error deliberately by deleting the ESLint disable comment at the bottom of `mobile/App.tsx` and verified that the commit was aborted

### Confirmation of Change 

| Creating a lint error deliberately |
| --- |
| image |


typokign (dismissed):
Looks fine, but is there any reason this is just for mobile/ and not web/ too?

dacruz21:
Also, is there any way we could do this with a [pre-push](https://www.git-scm.com/docs/githooks#_pre_push) hook instead?

justinyaodu:
> Looks fine, but is there any reason this is just for mobile/ and not web/ too?

Thanks for catching that. I didn't realize that there were two separate directories. My latest commit adds the Git hook to the web/ directory, and I tested it the same way I tested my other commit.

> Also, is there any way we could do this with a pre-push hook instead?

Good question. I used a pre-commit hook because that's what I saw on the different websites I looked at, but it should be possible to use a pre-push hook too. However, it might take a bit of effort to adapt the current script for pre-push instead of pre-commit, since pre-commit assumes that the relevant files are present in the working directory. For pre-push, this might not be the case if the branch being pushed isn't the current branch. I think the script would have to be modified to read the ref that's being pushed, check it out, run the lint commands, and go back to the previous HEAD (to avoid confusing the user with a branch switch). Do you feel that a pre-push hook would be more appropriate or convenient? I can consider that for a future version :slightly_smiling_face:

dacruz21:
Yeah I only bring up pre-push vs. pre-commit since our team tends to make a lot of commits before opening a PR. They can always just turn off hooks but I figured pre-push would be more convenient. Is that out of scope for this PR though?

Anshul-Birla:
> Yeah I only bring up pre-push vs. pre-commit since our team tends to make a lot of commits before opening a PR. They can always just turn off hooks but I figured pre-push would be more convenient. Is that out of scope for this PR though?

I second what David says. Linting on commit seems like a roadblock when you just made a commit with a solution that "just works" for rn and are going to clean it up later.

justinyaodu:
That rationale makes sense to me. I probably won't have time in the near future to make this into a pre-push hook, but it should be possible to continue using a workflow with frequent commits by using `git commit --no-verify` to skip the lint check. Does that sound like a suitable option? Alternatively, if the Git hook in its present state would cause more inconvenience than the value it provides, then we can close the PR. Up to you :slightly_smiling_face:

typokign (approved):
We can land this now if you'd like, but I do think a pre-push hook would be more appropriate.

justinyaodu:
I discussed this in more detail with David Hacker, and we agreed that having a mandatory lint check for every commit seemed excessive. Overall, I don't have any concerns about the quality of your team's work, so it might not be worth the inconvenience. Thank you for the feedback! I really appreciate it :slightly_smiling_face:

PR #31, opened 2021-05-04
Move home screen page numbers outside ScrollView

typokign:
### Administrative Info
Monday Board ID: 853642866

### Changes
At design team's request, move the page numbers outside of the
ScrollView, and replace the arrows with "Page M of N"

### Confirmation of Change 
Screenshot_20210503-190550

### DIFF
--- a/mobile/assets/i18n/en.json
+++ b/mobile/assets/i18n/en.json
@@ -12,5 +12,6 @@
     "explore": "Explore",
     "learn": "Learn",
     "liveChat" : "Live chat is coming soon",
-    "untilThen": "Until then..."
+    "untilThen": "Until then...",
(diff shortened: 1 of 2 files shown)


aksharans (changes requested):
Looks good, but the only thing is that when there are no results, the page numbers jump to the top and shows page 1 of 0:
   
image


I think this is because the no results screen is in the home page, so when there are no books, the paginated booklist scrollview becomes a null component, and the page number view renders then the no results renders.  

Could a fix be bookPages.length !== 0 && {View for pagesMofN} ?

dacruz21:
Good catch, that's exactly what I'll do :)

aksharans (approved):
LGTM

allisonbhavsar (approved):
Looks great!

PR #36, opened 2021-05-10
Admin API

k6han:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Added admin API (Create, Get, Update, Delete), along with new admin database

### Testing
- Added new file internal/controllers/admin_test.go,
- Updated book_test.go with authenticated requests

### Confirmation of Change 
Run make test in the backend directory


backend/internal/controllers/admin.go
```diff
@@ -0,0 +1,271 @@
...
+
+    // Create new admin in database
+    err = c.Admins.CreateAdmin(req.Context(), admin)
+    if err != nil {
+        writeResponse(rw, http.StatusInternalServerError, "error")
+        fmt.Println(err)
+        return
+    }
+
+    writeResponse(rw, http.StatusOK, cuid)
```
dacruz21: Let's write the entire `admin` object back to the response

backend/internal/controllers/admin.go
```diff
@@ -0,0 +1,271 @@
...
+
+    // Check auth token
+    uid, ok := req.Context().Value("user").(string)
+	if !ok {
+		writeResponse(rw, http.StatusInternalServerError, "error")
+		fmt.Println("unable to get user from request context")
+		return
+	}
+
+    // Check if admin
```
dacruz21: CanManageUsers

backend/internal/controllers/admin.go
```diff
@@ -0,0 +1,271 @@
...
+        return
+    }
+
+    // Delete the admin
+    if err = c.Admins.RemoveAdmin(req.Context(), adminID); err != nil {
+        writeResponse(rw, http.StatusInternalServerError, "error")
+        fmt.Println(err)
+    }
+
+    writeResponse(rw, http.StatusOK, "deleted")
```
dacruz21: Minor nit, let's do http.StatusNoContent and no response body

backend/internal/controllers/book.go
```diff
@@ -51,8 +53,29 @@ func (c *BookController) GetBookDetails(rw http.ResponseWriter, req *http.Reques
...
+    isAdmin, _, canUploadBooks, _, err := c.Admins.FetchAdminPermissions(req.Context(), uid)
+    if err != nil {
+        writeResponse(rw, http.StatusInternalServerError, "error")
+        fmt.Println(err)
+        return
+    }
+    if !isAdmin || !canUploadBooks {
+        writeResponse(rw, http.StatusForbidden, "do not have permission")
+        return
+    }
```
dacruz21: I would extract all of this into middleware - let's make some kind of a `RequireAdmin()` or `RequirePermission(CanManageUsers)` middleware in internal/controllers/middleware, then add it to the POST/PATCH/DELETE book routes.

backend/internal/controllers/router.go
```diff
@@ -23,11 +23,13 @@ func GetRouter(authenticator auth.Authenticator) chi.Router {
 	bookDB := database.BookDatabase{Conn: dbConn}
 	imageDB := database.ImgDatabase{Conn: dbConn}
 	userDB := database.UserDatabase{Conn: dbConn}
+	adminDB := database.AdminDatabase{Conn: dbConn}
 
 	// Set up the controller, which receives and responds to HTTP requests
-	bookController := BookController{Books: bookDB}
+    bookController := BookController{Books: bookDB, Admins: adminDB}
```
dacruz21: Ideally you should be able to get rid of this field once the middleware is done.

backend/internal/database/user.go
```diff
@@ -15,7 +15,7 @@ type UserDatabase struct {
 
 // Adds a user to the database
 func (db *UserDatabase) CreateUser(ctx context.Context, user models.User) error {
-	_, err := db.Conn.Exec(ctx, "INSERT INTO users (id, name, email, in_san_diego) VALUES ($1, $2, $3, $4) RETURNING id", user.ID, user.Name, user.Email, user.InSanDiego)
+	_, err := db.Conn.Exec(ctx, "INSERT INTO users (id, name, email, in_san_diego) VALUES ($1, $2, $3, $4)", user.ID, user.Name, user.Email, user.InSanDiego)
```
dacruz21, line 18: oh nice catch :)

backend/migrations/20210412123445_admin_table.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+CREATE TABLE admins(
+    id    text PRIMARY KEY,
+    email text NOT NULL UNIQUE,
+    name text NOT NULL,
+    can_manage_users boolean,
+    can_upload_books boolean,
+    can_delete_books boolean,
+    is_primary_admin boolean
```
dacruz21, line 9: These should all be NOT NULL DEFAULT false

typokign (changes requested):
Looking pretty good! Let's definitely make a middleware for validating admin permissions though.

You could write one middleware for each permission, but since you're pretty much doing the same thing for each permission, I would try to combine them into one, something like:

```go
middleware.RequireAdminPermission(auth Authenticator, p Permission)
```

You can define an enum for all of the permissions like so (maybe in models/admin.go):

```go
type Permission int

const (
	CanManageUsers Permission = iota
	CanUploadBooks
	CanDeleteBooks
)
```

https://qvault.io/golang/golang-enum/

Enums in Go aren't really type safe so you'll still have to have a check for `if p != CanManageUsers && p != CanUploadBooks && p != CanDeleteBooks` - but at least it makes the middleware config a little more expressive.

dacruz21:
Oh damn, sorry, I think I might have said we don't need a CanEditBooks permission but I do see that on the mockup: 
image - can you add that please? We should also add checks that you can't set CanEditBooks or CanDeleteBooks on an admin that doesn't have CanUploadBooks set

backend/internal/auth/firebase.go
```diff
@@ -20,3 +23,20 @@ func (a FirebaseAuthenticator) VerifyToken(ctx context.Context, token string) (s
...
 }
+
+func (a FirebaseAuthenticator) GenerateToken(ctx context.Context, email string, pwd string) (string, error) {
+    // Populate struct with fields for user creation
+    var u *auth.UserToCreate
+    u = u.Email(email)
+    u = u.Password(pwd)
+
+    // Create User
+    ur, err := a.Client.CreateUser(ctx, u)
```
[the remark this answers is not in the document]
k6han: My bad, didn't catch the bad pointer.
Just switching it to value messed up the u.Email and u.Password calls, so I worked around it:
var user auth.UserToCreate
u := &user

Let me know if theres a better way to do this lol

backend/internal/models/admin.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+type CreateAdmin struct {
+	Name           string `json:"name"`
+	Email          string `json:"email"`
+	Password       string `json:"password"`
+	CanManageUsers bool   `json:"can_manage_users"`
+	CanUploadBooks bool   `json:"can_upload_books"`
+	CanDeleteBooks bool   `json:"can_delete_books"`
+	IsPrimaryAdmin bool   `json:"is_primary_admin"`
```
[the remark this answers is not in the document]
k6han: This field is used for Update and Delete, to make sure that we cannot delete/update the primary admin. I could add this to the middleware, though.

Currently it is impossible to create more primary admins through the API, the database call will ignore that field and fill in false

backend/internal/auth/firebase.go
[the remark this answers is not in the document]
dacruz21: Yeah that's fine, there's really no reason to call `.Email()` and `.Password()` on a pointer instead of a value, so you could just change `u` to an `auth.UserToCreate` and only get its address on line 34 with `&u` but it works just the same

backend/internal/models/admin.go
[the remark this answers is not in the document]
dacruz21: > Currently it is impossible to create more primary admins through the API, the database call will ignore that field and fill in false

Ah good, thanks.

> This field is used for Update and Delete, to make sure that we cannot delete/update the primary admin

Meaning you trust the user input to determine if the admin is primary? I think it would be far safer to double-check ourselves.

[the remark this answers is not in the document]
k6han: Currently, Update and Delete take in an id to delete (admins/{id}). We then run getAdminByID, to check to see if the admin to update/delete even exists. If the admin does exist, then we check the isPrimaryAdmin field of the admin from the database, to see if we can proceed with the update/delete.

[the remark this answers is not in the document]
dacruz21: Got it, that sounds good - I don't think you need this field on the JSON body then, right?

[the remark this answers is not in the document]
k6han: Yeah I think we can remove the field on JSON.

Also how do we want to enforce that an admin cannot have Edit or Delete book permission without Create book permission? I was thinking of just doing something like 
editBook <- createBook && editBook
Do we want to report an error instead?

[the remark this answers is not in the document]
dacruz21: Yeah, I think it would be better to throw an error if the permissions get out of sync.

dacruz21:
I still think we should check permissions in a new middleware - thoughts?

k6han:
> I still think we should check permissions in a new middleware - thoughts?
Yeah you're right lol
I just made a bunch of unnecessary work for myself let me fix that

backend/internal/controllers/book.go
```diff
@@ -11,7 +11,8 @@ import (
 )
 
 type BookController struct {
-	Books database.BookDatabase
+	Books  database.BookDatabase
+	Admins database.AdminDatabase
```
dacruz21, line 15: This is unused now with the middleware :D

backend/internal/controllers/middleware/requirePermission.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/TritonSE/words-alive/internal/database"
+	"github.com/TritonSE/words-alive/internal/models"
+)
+
+// HTTP middleware that validates the Authorization header against the provided authenticator
+func RequirePermission(adminDB database.AdminDatabase, p models.Permission) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
```
dacruz21, line 13: Let's actually do the check that the enum is valid here, before the return statement. If it fails, go ahead and just `panic()`. That way if we've misconfigured any of the route handlers the backend won't even start, making it more likely we catch this in dev.

backend/internal/controllers/router.go
```diff
@@ -42,25 +45,31 @@ func GetRouter(authenticator auth.Authenticator) chi.Router {
...
+		r.With(middleware.RequireAuth(authenticator)).
+			With(middleware.RequirePermission(adminDB, models.CanDeleteBooks)).Delete("/{id}/{lang}", bookController.DeleteBookDetail)
 
-		r.Patch("/{id}", bookController.UpdateBook)
+		r.With(middleware.RequireAuth(authenticator)).
+			With(middleware.RequirePermission(adminDB, models.CanEditBooks)).Patch("/{id}", bookController.UpdateBook)
 
-		r.Patch("/{id}/{lang}", bookController.UpdateBookDetails)
+		r.With(middleware.RequireAuth(authenticator)).
+			With(middleware.RequirePermission(adminDB, models.CanEditBooks)).Patch("/{id}/{lang}", bookController.UpdateBookDetails)
```
dacruz21, line 68: Small thing, `r.With()` is variadic so you can clean this up by combining the RequireAuth and RequirePermission middlewares.

backend/internal/controllers/router.go
```diff
@@ -70,7 +79,16 @@ func GetRouter(authenticator auth.Authenticator) chi.Router {
...
+	r.With(middleware.RequireAuth(authenticator)).
+		With(middleware.RequirePermission(adminDB, models.CanManageUsers)).Post("/admins", adminController.CreateAdmin)
+	r.With(middleware.RequireAuth(authenticator)).
+		With(middleware.RequirePermission(adminDB, models.CanManageUsers)).Get("/admins", adminController.GetAdminList)
+	r.With(middleware.RequireAuth(authenticator)).
+		With(middleware.RequirePermission(adminDB, models.CanManageUsers)).Get("/admins/{id}", adminController.GetAdminByID)
+	r.With(middleware.RequireAuth(authenticator)).
+		With(middleware.RequirePermission(adminDB, models.CanManageUsers)).Patch("/admins/{id}", adminController.UpdateAdmin)
+	r.With(middleware.RequireAuth(authenticator)).
+		With(middleware.RequirePermission(adminDB, models.CanManageUsers)).Delete("/admins/{id}", adminController.DeleteAdmin)
```
dacruz21, line 91: ^

backend/internal/models/admin.go
```diff
@@ -0,0 +1,51 @@
...
+type CreateAdmin struct {
+	Name           string `json:"name"`
+	Email          string `json:"email"`
+	Password       string `json:"password"`
+	CanManageUsers bool   `json:"can_manage_users"`
+	CanUploadBooks bool   `json:"can_upload_books"`
+	CanEditBooks   bool   `json:"can_edit_books"`
+	CanDeleteBooks bool   `json:"can_delete_books"`
+	IsPrimaryAdmin bool
+}
```
dacruz21, line 25: I still think there's some confusion here on `IsPrimaryAdmin`. To clarify:

- We do want to show if an admin is primary in `GET /admins`, so IsPrimaryAdmin should have a json: tag in Admin
- We do _not_ want to allow the is_primary_admin field in PATCH requests, so the IsPrimaryAdmin field should be deleted entirely from CreateAdmin. AFAIK this struct is not being used for anything other than parsing API requests, and since IsPrimaryAdmin should never be on an API request, it shouldn't be on the CreateAdmin struct.

Also, deleting the json: tag doesn't stop a field from being marshalled/unmarshalled to/from JSON - it will just use the name of the field as the key. We only specify the json: tags to change format - from CanManageUsers to can_manage_users for example.

typokign (changes requested):
Middleware looks great, thanks. Just a few more small things then I think this will be g2g

typokign (dismissed):
Looks great, great work!!

backend/internal/controllers/admin.go
```diff
@@ -0,0 +1,199 @@
...
+	if err := json.NewDecoder(req.Body).Decode(&cadmin); err != nil {
+		writeResponse(rw, http.StatusBadRequest, "bad input!")
+		fmt.Println(err)
+		return
+	}
+
+	// Check for duplicate admin email
+	dupAdmin, err := c.Admins.FetchAdminByEmail(req.Context(), cadmin.Email)
+	if err != nil {
+		writeResponse(rw, http.StatusInternalServerError, "error")
```
dacruz21, line 36: 400 please :)

dacruz21: Oh wait oops - Aksharan was hitting this 500 and I thought it was because you were returning a 500 when there's a duplicate, misread this totally. Let me look and see if I can figure out why this was actually being hit

dacruz21: Ok never mind, this looks good. I wrote a test for duplicate email addresses, can you apply this patch please?

```patch
diff --git a/backend/internal/controllers/admin_test.go b/backend/internal/controllers/admin_test.go
index ff52423..5d5d0c6 100644
--- a/backend/internal/controllers/admin_test.go
+++ b/backend/internal/controllers/admin_test.go
@@ -16,7 +16,7 @@ func TestPostAdminUnauthorized(t *testing.T) {
 	fmt.Print("\n================ START ADMIN TESTS ================\n")
 	fmt.Print("\n---------------- CREATE ADMIN TESTS ----------------\n")

-	body := `{"email": "admin1@test.com", "name": "adam",
+	body := `{"email": "admin1@test.com", "password": "p@ssw0rd!", name": "adam",
     "can_manage_users": false, "can_upload_books": true, "can_delete_books" : true}`
 	// not passing token
 	testutils.MakeAuthenticatedRequest(t, "POST", ts.URL+"/admins", body,
@@ -27,7 +27,7 @@ func TestPostAdminUnauthorized(t *testing.T) {
 func TestPostAdminConflictingPerm(t *testing.T) {
 	var response string

-	body := `{"email": "admin1@test.com", "name": "adam",
+	body := `{"email": "admin1@test.com", "password": "p@ssw0rd!", "name": "adam",
     "can_manage_users": true, "can_upload_books": false,
     "can_edit_books": true, "can_delete_books" : false}`
 	testutils.MakeAuthenticatedRequest(t, "POST", ts.URL+"/admins", body,
@@ -40,7 +40,7 @@ func TestPostAdminConflictingPerm(t *testing.T) {
 func TestPostAdmin(t *testing.T) {
 	var response models.Admin

-	body := `{"email": "admin1@test.com", "name": "adam",
+	body := `{"email": "admin1@test.com", "password": "p@ssw0rd!", "name": "adam",
     "can_manage_users": true, "can_upload_books": true, "can_edit_books": true,
     "can_delete_books" : false}`
 	testutils.MakeAuthenticatedRequest(t, "POST", ts.URL+"/admins", body,
@@ -48,7 +48,7 @@ func TestPostAdmin(t *testing.T) {

 	require.Equal(t, "admin1", response.ID)

-	body = `{"email": "intern1@test.com", "name": "ian",
+	body = `{"email": "intern1@test.com", "password": "p@ssw0rd!", "name": "ian",
     "can_manage_users": false, "can_upload_books": false, "can_delete_books" : false}`

 	testutils.MakeAuthenticatedRequest(t, "POST", ts.URL+"/admins", body,
@@ -57,6 +57,18 @@ func TestPostAdmin(t *testing.T) {
 	require.Equal(t, "intern1", response.ID)
 }

+func TestPostAdminDuplicateEmail(t *testing.T) {
+	body := `{"email": "admin1@test.com", "password": "p@ssw0rd!", "name": "adam",
+    "can_manage_users": true, "can_upload_books": true, "can_edit_books": true,
+    "can_delete_books" : false}`
+
+	var res string
+	testutils.MakeAuthenticatedRequest(t, "POST", ts.URL+"/admins", body,
+		http.StatusBadRequest, &res, "test-token-primary")
+
+	require.Equal(t, "duplicate email", res)
+}
+
 // Test list admins without auth
 func TestListAdminsUnauthorized(t *testing.T) {
 	fmt.Print("\n---------------- GET ADMIN TESTS ----------------\n")

```

typokign (approved):

k6han:
@dacruz21 Any idea whats happening here?
I only changed the test file, and "make test" runs fine for me

dacruz21:
@k6han The monday bot? No clue, I'll turn off the merge restriction and report it to the awesome-tool team.

wes-chen:
The awesome tool check passes now - if you hit this error again please let me know. 

I suspect the tool was picking up the second colon on "Make sure that your..." , so the fix is to just delete the line. Every time the PR body is edited awesome-tool will retrigger. Sorry for the inconvenience, if you can reproduce it reliably please lmk

RohithKasar (approved):
Really nice work, looks good to me!

PR #48, opened 2021-06-01
Analytics API

allisonbhavsar:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Analytics API to manage daily book clicks
- Ability to increment the book click count for the current day
- Get book click counts in a specified day range from the current day [1, 366]

### Testing
- Run tests in book_test.go

### Confirmation of Change 
- curl -X PUT "http://localhost:8080/analytics/<book_id>/inc"
- curl -X GET -H "Authorization: Bearer test-token-<admin_id>" "http://localhost:8080/analytics/<book_id>?range=<[1, 366]>" (admin_id must have can_access_analytics permission)


backend/migrations/20210506010327_create_analytics_table.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE book_analytics (
+    id    text PRIMARY KEY,
+    clicks integer[],
```
dacruz21, line 5: I would add a `NOT NULL DEFAULT array_fill(0, array[366])` here, and remove the `(clicks) / array_fill` from your insert query below. That way we can ensure that the array never gets deleted somehow.

backend/internal/database/book.go
```diff
@@ -246,3 +254,90 @@ func (db *BookDatabase) UpdateBookDetails(ctx context.Context, id string,
...
+		var prevYear = currTime.Year() - 1
+		var lastDay = time.Date(prevYear, time.December, 31, 0, 0, 0, 0, time.Local).YearDay()
+		arg1 = lastDay - (dRange - dayInYear) + 1
+		arg2 = lastDay
+		arg3 = dayInYear
+	} else {
+		arg1 = dayInYear - dRange + 1
+		arg2 = dayInYear
+		arg3 = 0
+	}
```
dacruz21, line 318: Definitely the most annoying part of this API :P Can you please:

- Rename arg1/2/3 to something a little more clear? (Like `prevYearRangeStart` or something like that?)
- Add some comments explaining how this all works
- Maybe a couple of examples in comments on how this works with different date ranges?

backend/internal/database/book.go
```diff
@@ -246,3 +254,90 @@ func (db *BookDatabase) UpdateBookDetails(ctx context.Context, id string,
 	return updatedBookDetails, err
 
 }
+
+/*
+ * Increments current day's clicks count in the book_analytics table given a book id
+ */
+ func (db *BookDatabase) UpdateBookAnalytics(ctx context.Context, id string) error {
```
dacruz21, line 261: I would rename this to `IncAnalyticsCounter` or something like that.

backend/go.mod
```diff
@@ -4,18 +4,24 @@ go 1.15
...
 	github.com/jackc/pgx/v4 v4.10.1
-	github.com/lib/pq v1.9.0 // indirect
+	github.com/lib/pq v1.10.1 // indirect
+	github.com/mattn/go-sqlite3 v1.14.7 // indirect
 	github.com/pkg/errors v0.9.1
 	github.com/pressly/goose v2.7.0+incompatible
 	github.com/shopspring/decimal v1.2.0 // indirect
 	github.com/stretchr/testify v1.6.1
+	github.com/ziutek/mymysql v1.5.4 // indirect
+	golang.org/x/crypto v0.0.0-20210503195802-e9a32991a82e // indirect
```
dacruz21, line 24: I don't think you added any dependencies in this PR, can you revert this file and go.sum? Correct me if I'm wrong :)

backend/internal/controllers/book.go
```diff
@@ -209,3 +210,62 @@ func (c *BookController) UpdateBookDetails(rw http.ResponseWriter, req *http.Req
...
+
+	if numDays, err := strconv.Atoi(dayRange); err != nil {
+		writeResponse(rw, http.StatusBadRequest, "could not parse range")
+		return
+	} else if numDays < 1 || numDays > 366 {
+		writeResponse(rw, http.StatusBadRequest, "Range out of bounds")
+		return
+	} else {
+		dRange = numDays
+	}
```
dacruz21, line 255: You can change `if numDays, err :=` to `if dRange, err :=` in your if statement, and get rid of the `else` block - since dRange is already defined it won't be scoped to just the if statement.

You could also then split the `else if` into a separate `if` statement using dRange instead of numDays

Actually, I would probably just rename dRange to numDays, I think that's a better variable name.

typokign (changes requested):
Great! Really close and I think your year wrap-around implementation is solid. A couple comments below. If you want to keep building on this, here's where we want to take the analytics API next:

- Call PUT /inc from BookScreen.tsx in mobile, so we actually start gathering these statistics
- Add GET /books/popular, which returns the 5 books with the most total clicks over the last 7 days, sorted by that click count (and optionally, integrate this into the frontend, to build the Popular Books section of the home screen)
- Batch PUT /inc calls - I think in the long run this API is going to hammer the database. If you want to learn some more advanced Go (goroutines, channels, etc.) I can give you some ideas on how to move the incrementing to a separate thread and perform multiple increments at once.

Totally optional though, it is the end of the year after all :)

backend/internal/controllers/book.go
```diff
@@ -209,3 +210,62 @@ func (c *BookController) UpdateBookDetails(rw http.ResponseWriter, req *http.Req
...
+
+	err := c.Books.UpdateBookAnalytics(req.Context(), bookID)
+
+	if err != nil {
+		writeResponse(rw, http.StatusInternalServerError, "error")
+		return
+	}
+
+	writeResponse(rw, http.StatusNoContent, nil)
```
dacruz21, line 232: Oh, also, I think StatusOK makes more sense here than NoContent - you don't always have to return 204 if there's nothing in the body, I've really only seen it as a convention for DELETE routes.

dacruz21:
Just tested this locally, works well, didn't really bother testing the year wrap around extensively but the logic looks sound. One issue though, you never create a book_analytics row for new books besides the database migration. Keep in mind that the migration only runs once, so any books added after we release this PR to prod will never have a book_analytics row. I would add a `INSERT INTO book_analytics` query in POST /books, though you could also do it on the first PUT /inc if you really wanted to. (I think on POST would be cleaner)

Also, I would maybe consider splitting the /analytics API into its own controller and database struct? Not super necessary quite yet but we will definitely be adding more to this API and the books API later and I think things might get bloated. Up to you if you have time to do that in this PR or not.

backend/internal/database/book.go
```diff
@@ -129,6 +130,13 @@ func (db *BookDatabase) InsertBook(ctx context.Context,
 		return nil, errors.Wrap(err, "error on INSERT INTO books in InsertBook")
 	}
 
+	query = "INSERT INTO book_analytics (id) VALUES($1);"
```
allisonbhavsar, line 133: @dacruz21 Shouldn't this take care of updating book_analytics whenever new books are added?

typokign (approved):
Oh, yep, not sure what was going on but it works now. :thinking:

PR #59, opened 2021-11-05
Favorite Books API

ecflores30:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Added favorite books table and made migrations
- Implemented functions to get, add to, and delete from favorite books table, returning proper HTTP status response
- Connected API routes to appropriate controller functions

### Testing
- Wrote unit tests to verify functionality of getting, adding, and deleting favorite books for authenticated users
- Ran `make test` and confirmed that all unit tests passes

### Confirmation of Change 

Favorite Books Unit Tests
image

### DIFF
--- a/backend/internal/controllers/book.go
+++ b/backend/internal/controllers/book.go
@@ -5,15 +5,18 @@ import (
 	"log"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/go-chi/chi"
 
+	"github.com/TritonSE/words-alive/internal/auth"
 	"github.com/TritonSE/words-alive/internal/database"
 	"github.com/TritonSE/words-alive/internal/models"
 )
 
 type BookController struct {
 	Books database.BookDatabase
+	Auth  auth.Authenticator
 }
 
 // Fetches a list of all books sorted by title for the main page (/books)
@@ -30,13 +33,15 @@ func (c *BookController) GetBookList(rw http.ResponseWriter, req *http.Request) 
 
 // GetBook fetches a single book from the database by its ID
 func (c *BookController) GetBook(rw http.ResponseWriter, req *http.Request) {
-	id := chi.URLParam(req, "id")
-	if id == "" {
+
+	var bookID string = chi.URLParam(req, "id")
+
+	if bookID == "" {
 		writeResponse(rw, http.StatusBadRequest, "invalid ID")
 		return
 	}
 
-	book, err := c.Books.FetchBook(req.Context(), id)
+	book, err := c.Books.FetchBook(req.Context(), bookID)
 	if err != nil {
 		log.Println(err)
 		writeResponse(rw, http.StatusInternalServerError, "error")
@@ -45,6 +50,32 @@ func (c *BookController) GetBook(rw http.ResponseWriter, req *http.Request) {
 	if book == nil {
 		writeResponse(rw, http.StatusNotFound, "book not found")
 		return
+	}
+
+	// if user is authenticated, check if the book is favorited
+	authHeader := req.Header.Get("Authorization")
+	if authHeader != "" {
+
+		if !strings.HasPrefix(authHeader, "Bearer ") {
+			writeResponse(rw, http.StatusBadRequest, "Unauthenticated requests should unset the header")
+			return
+		}
+
+		tokenString := strings.TrimPrefix(authHeader, "Bearer ")
+		userID, ok := c.Auth.VerifyToken(req.Context(), tokenString)
+		if !ok {
+			writeResponse(rw, http.StatusForbidden, "Token was invalid")
+			return
+		}
+
+		// check if the book is favorited if userID not empty
+		book.Favorite, err = c.Books.FetchFavoritedBook(req.Context(), userID, bookID)
+
+		if err != nil {
+			log.Println(err)
+			writeResponse(rw, http.StatusInternalServerError, "error")
+			return
+		}
 	}
 
 	writeResponse(rw, http.StatusOK, book)
@@ -245,7 +276,7 @@ func (c *BookController) UpdateBookDetails(rw http.ResponseWriter, req *http.Req
 
 // increments a book's click count for the current day (/analytics/{id}/inc)
 func (c *BookController) UpdateBookClicks(rw http.ResponseWriter, req *http.Request) {
-	var bookID string = chi.URLParam(req, "id")
+	bookID := chi.URLParam(req, "id")
 
 	analytics, _ := c.Books.FetchBookAnalytics(req.Context(), bookID, 1)
 
@@ -263,7 +294,6 @@ func (c *BookController) UpdateBookClicks(rw http.ResponseWriter, req *http.Requ
 	}
 
 	writeResponse(rw, http.StatusOK, nil)
-
 }
 
 // fetches daily book clicks in a given range (/analytics/{id}?range=<days>)
@@ -300,7 +330,6 @@ func (c *BookController) GetBookClicks(rw http.ResponseWriter, req *http.Request
 	}
 
 	writeResponse(rw, http.StatusOK, bookAnalytics)
-
 }
 
 // Fetches daily book clicks in a given range from today (/analytics?range=<days>)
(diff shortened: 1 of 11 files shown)


backend/internal/controllers/book.go
```diff
@@ -285,3 +286,75 @@ func (c *BookController) GetBookClicks(rw http.ResponseWriter, req *http.Request
...
+
+	writeResponse(rw, http.StatusOK, favorites)
+
+}
+
+// creates a new entry in the favorite books table (/books/favorites/{id})
+func (c *BookController) AddToFavorites(rw http.ResponseWriter, req *http.Request) {
+
+	userID, ok := req.Context().Value("user").(string)
+	var bookID string = chi.URLParam(req, "id")
```
dacruz21, line 315: Can we validate the book ID? Right now this will 500 from the foreign key constraint but ideally we should return a 400 instead. You can check the error returned by InsertFavoriteBook or just try to query for the book with that ID before ever attempting an insert.

backend/internal/controllers/book_test.go
```diff
@@ -383,3 +383,75 @@ func TestGetAnalytics(t *testing.T) {
...
 }
+
+/*
+1. make unauthenticated request - should get HTTP 401 - COMPLETE
+2. make authenticated request - should get no favorite books
+3. favorite a book - COMPLETE
+4. get favorite books, make sure book from step 3 is in there
+5. unfavorite the book
+6. get favorite books, response should be empty again
+*/
```
dacruz21, line 394: this can be removed I think

backend/internal/database/book.go
```diff
@@ -357,3 +357,123 @@ func (db *BookDatabase) FetchBookAnalytics(ctx context.Context, id string, dRang
...
 
 }
+
+/*
+ * Returns an array containing the favorite books for the given userID
+ */
+func (db *BookDatabase) FetchFavorites(ctx context.Context, userID string) ([]models.Book, error) {
+
+	var favoriteBooks []models.Book
+	favoriteBooks = make([]models.Book, 0)
```
dacruz21, line 367: you can combine these into `favoriteBooks := make([]models.Book, 0)`

Actually, you shouldn't need to call make here at all, append will make the array if it's not constructed yet.

backend/internal/database/book.go
```diff
@@ -357,3 +357,123 @@ func (db *BookDatabase) FetchBookAnalytics(ctx context.Context, id string, dRang
...
+
+		if err2 != nil {
+			fmt.Print(err2)
+			return nil, errors.Wrap(err, "error on FetchBook from FetchFavorites")
+		}
+
+		// Add favorite book to array
+		favoriteBooks = append(favoriteBooks, *book)
+
+	}
```
dacruz21, line 403: This is needlessly [N+1](https://secure.phabricator.com/book/phabcontrib/article/n_plus_one/)

Can you combine this into the `SELECT FROM favorite_books` query by way of a join or `book.id IN` test?

backend/internal/database/book.go
```diff
@@ -357,3 +357,123 @@ func (db *BookDatabase) FetchBookAnalytics(ctx context.Context, id string, dRang
...
+
+	var query = "SELECT book_id FROM favorite_books WHERE user_id = $1 AND book_id = $2"
+
+	err := db.Conn.QueryRow(ctx, query, userID, bookID).Scan(&book_id)
+
+	if err != nil {
+		return "", errors.Wrap(err, "error on FetchFavoritedBook")
+	}
+
+	_, err2 := db.FetchBook(ctx, book_id)
```
dacruz21, line 430: These two queries can also be combined.

backend/internal/models/user.go
```diff
@@ -6,6 +6,7 @@ type User struct {
 	Name       string `json:"name"`
 	Email      string `json:"email"`
 	InSanDiego bool   `json:"in_san_diego"`
+	Favorites  []Book `json:"favorite_books"`
```
dacruz21, line 9: Let's not do this, this is the model for GET /users/:id. We don't need the favorites array anywhere other than the favorites API

typokign (changes requested):
Comments below. Can you also revert web/package-lock.json? Not sure why that changed.

backend/internal/database/book.go
```diff
@@ -361,135 +361,39 @@ func (db *BookDatabase) FetchBookAnalytics(ctx context.Context, id string, dRang
...
 
 	var favoriteBooks []models.Book
-	favoriteBooks = make([]models.Book, 0)
 
-	var query string = "SELECT book_id FROM favorite_books WHERE user_id = $1"
+	var query string = "SELECT books.id, title, author, image, created_at, " +
+		"array_remove(array_agg(lang), NULL) as languages " +
+		"FROM books LEFT JOIN book_contents ON books.id = book_contents.id " +
+		"JOIN favorite_books ON books.id = favorite_books.book_id " +
+		"WHERE favorite_books.user_id = $1 GROUP BY books.id ORDER BY books.title"
```
dacruz21, line 373: Maybe we should rebase this branch off of my refactor, and do this query:

```sql
SELECT id, title, author, image, languages, created_at 
FROM books
WHERE id IN (
  SELECT book_id FROM favorite_books WHERE user_id = $1
);
```

typokign (dismissed):

aksharans:
@dacruz21 for the user context in getBooks, we wouldn't necessarily need the ok check right since there wouldn't be an error if the user isn't authenticated, so do we just ignore that and only check if the userID isn't empty or would we need to have the ok check somewhere like add `&& ok` to `userID != ""`?

dacruz21:
The user context is only set on routes that use the requireAuth middleware, see here for its implementation: https://github.com/TritonSE/WA-Family-Literacy-Application/blob/main/backend/internal/controllers/middleware/requireAuth.go#L15-L31

You'll unfortunately need to just duplicate some of this code since the middleware will block all unauthenticated requests with a 401 (which we don't want, we want unauthenticated users to be able to access this route without the favorite bool).

Rough order of steps:

1. Check for the `Authorization` header
2. If unset, no problem, continue without the favorite bool
3. If set, make sure the header value starts with Bearer and strip it. If the header is set but is invalid it's fine to return a 400, since unauthenticated requests should unset the header entirely.
4. Validate the token with the authenticator. If invalid, fail with a 403.
5. If everything checks out extract the user ID and use it in your favorites logic.

backend/internal/models/book.go
```diff
@@ -29,6 +29,7 @@ type Book struct {
 	Image     *string   `json:"image"`      // image link
 	Languages []string  `json:"languages"`  // Array of languages (ISO 639-1)
 	CreatedAt time.Time `json:"created_at"` // Following ISO 8601
+	Favorited bool      `json:"favorited"`  // if book is favorited
```
dacruz21, line 32: small nit, move this above created_at and maybe just call it "favorite"?

backend/internal/controllers/book.go
```diff
@@ -47,6 +56,24 @@ func (c *BookController) GetBook(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	// userID empty if and only if signed-in user makes get book request
+	if userID != "" {
+
+		favorited_id, err := c.Books.FetchFavoritedBook(req.Context(), userID, bookID)
```
dacruz21, line 62: this should probably just return a (bool, error) -> true if bookID is favorited by userID and false if not

backend/internal/database/book.go
```diff
@@ -402,8 +402,9 @@ func (db *BookDatabase) FetchFavorites(ctx context.Context, userID string) ([]mo
  * Fetches a single favorited book for the user, if it is favorited
  */
 func (db *BookDatabase) FetchFavoritedBook(ctx context.Context,
-	userID string, bookID string) (string, error) {
+	userID string, bookID string) (bool, error) {
 
+	var favorited bool
 	var book_id string
 
 	query := "SELECT book_id FROM favorite_books JOIN books ON favorite_books.book_id = books.id " +
```
dacruz21, line 410: try `SELECT count(*) > 0 FROM favorite_books WHERE user_id=$1 AND book_id=$2`, we shouldn't need to join here since we're just checking the ID. Also there's no need to return the ID, we can just do a check if the row exists with count(*) > 0

typokign (approved):
idk if this counts since now I'm a committer but lgtm

lgtm

petabite (approved):
LGTM

PR #60, opened 2021-11-11
Realtime Chat API

petabite:
### Administrative Info
Monday Board ID: 853642866

### Changes
- Build chat API functionality using Firestore (live chat, rating/resolving a chat)
- Update the backend to save `canChat` custom claim to the Firebase Authentication user when an admin is created and updated (if an admins chat permission is updated, they must sign out and sign in again to refresh their token for the change to take effect)
- Update Firebase staging environment with new security rules to manage access to documents

#### Firestore Schemas

**ChatRoom**

```js
{
    createdAt: string, 
    rating: number, // user rating of chat
    resolved: boolean,
    user: string, // name of user who initiated the chat (for displaying on frontend, saves a read operation)
    userId: string // id of user who initiated the chat (used by the Firestore security rules for access control) 
    messages: Subcollection<Message> (see below)
}
```

**Message**

```js
{
    from: string, // user who sent the message (used for rendering correct chat bubble for frontend implementation in the future)
    sentAt: string,
    text: string
}
```

### Testing
- built a rough frontend on web and mobile to test functionality
- made sure chat permission was correctly added/removed on admin create/update
- made sure admins with chat permission have access to all chat rooms
- made sure only logged in users could see their own chat rooms (frontend guard not implemented yet for guest users)

### Confirmation of Change 
image

### DIFF
--- a/backend/internal/auth/auth.go
+++ b/backend/internal/auth/auth.go
@@ -14,5 +14,5 @@ type Authenticator interface {
 	VerifyToken(ctx context.Context, token string) (string, bool)
 	CreateUser(ctx context.Context, email string, pwd string) (string, error)
 	DeleteUser(ctx context.Context, id string) error
- 	SetCustomUserClaims(ctx context.Context, id string, claims map[string]interface{}) error
+	SetCustomUserClaims(ctx context.Context, id string, claims map[string]interface{}) error
 }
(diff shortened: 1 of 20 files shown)


typokign (dismissed):
Seems to work! I noticed some messages showing out of order but I think that's just a UI issue.

typokign (approved):

aksharans (approved):
LGTM

PR #62, opened 2021-11-12
Refactor books API

typokign:
### Administrative Info
Monday Board ID: 853642866

### Changes
What changes did you make?
- Renamed `books` table to `book_info`
- Added a view `books` to replace the table, which pulls in languages from `book_contents` automatically, without having to do a join on every query
- Updated GET /books/id to return a 404 if the book is not found

### Testing
- Tests pass


ecflores30 (approved):
Looks awesome!

aksharans (approved):
LGTM

PR #69, opened 2021-12-05
Use react-native-easy-markdown to fix rendering issues

typokign:
### Administrative Info
Monday Board ID: 853642866


petabite (approved):
looks good on my iphone and local simulator!

